PR #2, opened 2016-10-26
Replace package vendoring with current name

jeffandersen:
This fix will make building torus-cli a lot easier (aka, make it actually work).


ianlivingstone:
LGTM merge on 🍏

PR #4, opened 2016-10-27
Add confirm dialogue for destructive commands

jeffandersen:
This prevents users from accidentally destroying something unintended.

Right now our only destructive command is `unset`

### Basic usage:

```
> torus unset testunset
You are about to unset "/jeff/all-projects/dev-jeff/default/*/*". This cannot be undone.
✗ Do you wish to continue? [Y/n] n

> torus unset testunset
You are about to unset "/jeff/all-projects/dev-jeff/default/*/*". This cannot be undone.
✔ Do you wish to continue? [Y/n] Y
Credentials retrieved
Keypairs retrieved
Encrypting key retrieved
Credential encrypted
Completed Operation

Credential testunset has been unset at /jeff/all-projects/dev-jeff/default/*/*/testunset.
```
### Yes flag
```
> torus set testunset 1234 -y
Credentials retrieved
Keypairs retrieved
Encrypting key retrieved
Credential encrypted
Completed Operation

Credential testunset has been set at /jeff/all-projects/dev-jeff/default/*/*/testunset
```
### Auto confirm pref
```
> torus prefs set core.auto_confirm true
Preferences updated.

> torus set testunset 1234
Credentials retrieved
Keypairs retrieved
Encrypting key retrieved
Credential encrypted
Completed Operation

Credential testunset has been set at /jeff/all-projects/dev-jeff/default/*/*/testunset
```

### DIFF
--- a/cmd/set.go
+++ b/cmd/set.go
@@ -78,7 +78,7 @@ func setCmd(ctx *cli.Context) error {
 	return nil
 }
 
-func setCredential(ctx *cli.Context, nameOrPath string, valueMaker func() *apitypes.CredentialValue) (*apitypes.CredentialEnvelope, error) {
+func determineCredential(ctx *cli.Context, nameOrPath string) (*pathexp.PathExp, *string, error) {
 	// First try and use the cli args as a full path. it should override any
 	// options.
 	idx := strings.LastIndex(nameOrPath, "/")
@@ -92,14 +92,14 @@ func setCredential(ctx *cli.Context, nameOrPath string, valueMaker func() *apity
 		path := nameOrPath[:idx]
 		pe, err = pathexp.Parse(path)
 		if err != nil {
-			return nil, errs.NewExitError(err.Error())
+			return nil, nil, errs.NewExitError(err.Error())
 		}
 	} else {
 		// Falling back to flags. do the expensive population of the user flag now,
 		// and see if any required flags (all of them) are missing.
 		err := chain(setUserEnv, checkRequiredFlags)(ctx)
 		if err != nil {
-			return nil, err
+			return nil, nil, err
 		}
 
 		pe, err = pathexp.New(ctx.String("org"), ctx.String("project"),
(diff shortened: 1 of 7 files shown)


cmd/set.go
```diff
@@ -78,7 +78,7 @@ func setCmd(ctx *cli.Context) error {
 	return nil
 }
 
-func setCredential(ctx *cli.Context, nameOrPath string, valueMaker func() *apitypes.CredentialValue) (*apitypes.CredentialEnvelope, error) {
+func determineCredential(ctx *cli.Context, nameOrPath string) (*pathexp.PathExp, *string, error) {
```
jeffandersen: Doesn't really matter, just a cleaner return I guess

cmd/unset.go
```diff
@@ -35,7 +35,20 @@ func unsetCmd(ctx *cli.Context) error {
...
 
-	cred, err := setCredential(ctx, args[0], func() *apitypes.CredentialValue {
+	pathexp, _, err := determineCredential(ctx, args[0])
+	if err != nil {
+		return errs.NewErrorExitError("Could not unset credential", err)
+	}
+
+	preamble := fmt.Sprintf("You are about to unset \"%s\". This cannot be undone.", pathexp.String())
+
+	abort := ConfirmDialogue(ctx, nil, &preamble)
```
ianlivingstone: since `ConfirmDialog` retuns an error, we should call it `err` or `abortErr`. Perhaps a better interface would be `bool, err` for `ConfirmDialog` since a CTRL+C and a `Yes` to confirm are different.

You can see a world where we'd want to print out, `Aborted.` or similar.

jeffandersen: `bool, err` won't work because the abstraction is meant to construct the exit error.

ianlivingstone:
I'm going to open an issue for the prompt behaviour for us to discuss and tackle outside this pull request.

LGTM on 🍏

PR #6, opened 2016-10-28
Machines Create

ianlivingstone:
Introducing the ability to create a machine!

**UI**

``` bash
Ians-MacBook-Pro-2:torus-cli ianlivingstone$ ./torus machines create --help
NAME:
   torus machines create - Create a machine for an organization

USAGE:
   torus machines create [command options]

OPTIONS:
   --org ORG, -o ORG     the org the machine will belong too
   --team TEAM, -t TEAM  the team the machine will belong too

Ians-MacBook-Pro-2:torus-cli ianlivingstone$ ./torus machines create
✔ Org name: test
✔ Select Machine Team: api
✔ Enter machine name: api-98g4vw63

Creating machine "api-98g4vw63"
Generating machine token
Generating token keypairs
Uploading token keypairs
Creating keyring memberships for token
Uploading keyring memberships
Machine created

You will only be shown the secret once, please keep it safe.

Machine ID:           04bm8jj0dbxbbtgn7dey2b7tewt82
Machine Token ID:     04c5e58npy50g8meprw5etk514ha2
Machine Token Secret: TlxwEbG3Ol7K9PLSUCSVtAHk
```

**Changes**

I've tried to break the commits down as much as possible to be incremental and atomic for readability. I recommend reviewing by walking through each commit in linear order.

Each commit generally encapsulates the reasoning for the change, however, I'll cover the big changes here:
- `Session` now supports the concept of `identity` and `auth` objects in order to support machines which identify using the Machine object but authorize using a Machine Token object.
- I've fixed a few bugs with prompts (labelling, and not hiding errors) while also introducing a new prompt for selecting or creating teams.
- I've removed the dependency on the DB from the Crypto Engine. Instead, it uses the Session object for pulling the Master Key value from the User (or Machine) object. This was required to support machine creation where we now generate and encrypt keys on the behalf of another entity (machine creating a machine, or user creating a machine)
- I've add beginnings of support for Passphrase-Derived Public Key Authentication which will be used when we implement `machines login`.

### DIFF
--- a/api/teams.go
+++ b/api/teams.go
@@ -22,6 +22,30 @@ type TeamResult struct {
 	Body    *primitive.Team `json:"body"`
 }
 
+// List retrieves all teams for an org based on the filtered values
+func (t *TeamsClient) List(ctx context.Context, orgID *identity.ID, name, teamType string) ([]TeamResult, error) {
+	v := &url.Values{}
+
+	if orgID != nil {
+		v.Set("org_id", orgID.String())
+	}
+	if name != "" {
+		v.Set("name", name)
+	}
+	if teamType != "" {
+		v.Set("type", teamType)
+	}
+
+	req, _, err := t.client.NewRequest("GET", "/teams", v, nil, true)
+	if err != nil {
+		return nil, err
+	}
+
+	teams := []TeamResult{}
+	_, err = t.client.Do(ctx, req, &teams, nil, nil)
+	return teams, err
+}
+
 // GetByOrg retrieves all teams for an org id
 func (t *TeamsClient) GetByOrg(ctx context.Context, orgID *identity.ID) ([]TeamResult, error) {
 	v := &url.Values{}
@@ -54,20 +78,25 @@ func (t *TeamsClient) GetByName(ctx context.Context, orgID *identity.ID, name st
 }
 
 // Create performs a request to create a new team object
-func (t *TeamsClient) Create(ctx context.Context, orgID *identity.ID, name string) error {
+func (t *TeamsClient) Create(
+	ctx context.Context, orgID *identity.ID, name, teamType string) (*apitypes.Team, error) {
 	if orgID == nil {
-		return errors.New("invalid org")
+		return nil, errors.New("invalid org")
+	}
+
+	if teamType == "" {
+		teamType = primitive.UserTeam
 	}
 
 	teamBody := primitive.Team{
 		Name:     name,
 		OrgID:    orgID,
-		TeamType: "user",
+		TeamType: teamType,
 	}
 
 	ID, err := identity.NewMutable(&teamBody)
 	if err != nil {
-		return err
+		return nil, err
 	}
 
 	team := apitypes.Team{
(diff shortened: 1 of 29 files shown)


cmd/machines.go
```diff
@@ -0,0 +1,180 @@
...
+
+	args := ctx.Args()
+	name := ""
+	if len(args) > 0 {
+		name = args[0]
+	}
+
+	name, err = promptForMachineName(name, teamName)
+	if err != nil {
+		return handleSelectError(err, projectCreateFailed)
```
jeffandersen: Wrong error const here

cmd/machines.go
```diff
@@ -0,0 +1,180 @@
...
+	return machine, tokenSecret, nil
+}
+
+func promptForMachineName(providedName, teamName string) (string, error) {
+	defaultName, err := deriveMachineName(teamName)
+	if err != nil {
+		return "", errs.NewErrorExitError("Could not derive machine name", err)
+	}
+
+	name := ""
```
jeffandersen: `name := providedName`

ianlivingstone: Fixed 💯

cmd/machines.go
```diff
@@ -0,0 +1,186 @@
...
+		Name:     "machines",
+		Usage:    "Manage machine for an organization",
+		Category: "ORGANIZATIONS",
+		Subcommands: []cli.Command{
+			{
+				Name:  "create",
+				Usage: "Create a machine for an organization",
+				Flags: []cli.Flag{
+					orgFlag("the org the machine will belong too", false),
+					teamFlag("the team the machine will belong too", false),
```
jeffandersen: Two typos: `too` -> `to`

cmd/teams.go
```diff
@@ -158,16 +158,23 @@ func teamsListCmd(ctx *cli.Context) error {
...
-		teamType := ""
-		if t.Body.TeamType == primitive.SystemTeam {
-			teamType = "[system]"
+		displayTeamType := ""
+
+		switch teamType := t.Body.TeamType; teamType {
+		case primitive.SystemTeam:
+			displayTeamType = "[system]"
+			if t.Body.Name == "machine" {
+				displayTeamType += " (machine)"
```
ianlivingstone: to visually differentiate between system and machine teams (otherwise it runs together on the screen)

cmd/machines.go
```diff
@@ -0,0 +1,186 @@
...
+	}
+
+	machine, tokenSecret, err := createMachineByName(c, client, orgID, teamID, name)
+	if err != nil {
+		return err
+	}
+
+	fmt.Print("You will only be shown the secret once, please keep it safe.\n\n")
+
+	w := tabwriter.NewWriter(os.Stdout, 2, 0, 1, ' ', 0)
```
ianlivingstone: 💯

api/teams.go
```diff
@@ -22,6 +22,30 @@ type TeamResult struct {
 	Body    *primitive.Team `json:"body"`
 }
 
+// List retrieves all teams for an org based on the filtered values
+func (t *TeamsClient) List(ctx context.Context, orgID *identity.ID, name, teamType string) ([]TeamResult, error) {
+	v := &url.Values{}
+
+	if orgID != nil {
```
ianlivingstone: the end-point doesn't support it today, so I'd rather match what it supports.

Over time, I think these endpoints will fade away as we force more things into the larger data structures.

PR #7, opened 2016-10-30
Machines list

jeffandersen:
Branched from https://github.com/manifoldco/torus-cli/pull/6

Will continue to update this as code review happens for machines create.

Sample output:

```
> torus machines list --help
NAME:
   torus machines list - List machines for an organization

USAGE:
   torus machines list [command options]

OPTIONS:
   --org ORG, -o ORG     Org to the machine belongs to
   --team TEAM, -t TEAM  Team the machine belongs to
   --destroyed, -d       Display only destroyed
```

```
> torus machines list

ID                                   NAME                              STATE         TEAM                     CREATION DATE

04buv5v36xecb5r390btvevkh0k10        api-test-machines-00cjftz6        active        api-test-machines        2016-10-30T12:56:21Z
04bh45uqk811xqzr81jb6naw5mt02        api-test-machines-b5hryd28        active        api-test-machines        2016-10-30T12:56:35Z
04bhu01f40b94k8vw7drh0gr294xy        ok-8u6j7cwv                       active        ok                       2016-10-30T19:56:21Z
```

```
> torus machines list --destroyed

ID                                   NAME                              STATE            TEAM                     CREATION DATE

04bx7jwzw9c8wd6t0n7mqpxerqncu        api-test-machines-2apzzf71        destroyed        api-test-machines        2016-10-30T12:56:41Z

```

```
> torus machines list --team ok

ID                                   NAME               STATE         TEAM        CREATION DATE

04bhu01f40b94k8vw7drh0gr294xy        ok-8u6j7cwv        active        ok          2016-10-30T19:56:21Z
```

### DIFF
--- a/api/machines.go
+++ b/api/machines.go
@@ -3,6 +3,7 @@ package api
 import (
 	"context"
 	"crypto/rand"
+	"net/url"
 
 	"github.com/manifoldco/torus-cli/apitypes"
 	"github.com/manifoldco/torus-cli/base64"
@@ -55,4 +56,50 @@ func createTokenSecret() (*base64.Value, error) {
 	}
 
 	return base64.NewValue(value), nil
+}
+
+// List machines in the given org and state
+func (m *MachinesClient) List(ctx context.Context, orgID *identity.ID, state *string, name *string, teamID *identity.ID) ([]*apitypes.MachineSegment, error) {
+	v := &url.Values{}
+	if orgID != nil {
+		v.Add("org_id", (*orgID).String())
+	}
+	if state != nil {
+		v.Add("state", *state)
+	}
+	if teamID != nil {
+		v.Add("team_id", (*teamID).String())
+	}
+	if name != nil {
+		v.Add("name", *name)
+	}
+
+	req, reqID, err := m.client.NewRequest("GET", "/machines", v, nil, true)
+	if err != nil {
+		return nil, err
+	}
+
+	var results []*apitypes.MachineSegment
+	_, err = m.client.Do(ctx, req, &results, &reqID, nil)
+	if err != nil {
+		return nil, err
+	}
+
+	return results, nil
+}
+
+// Get machine by ID
+func (m *MachinesClient) Get(ctx context.Context, machineID *identity.ID) (*apitypes.MachineSegment, error) {
+	req, reqID, err := m.client.NewRequest("GET", "/machines/"+machineID.String(), nil, nil, true)
+	if err != nil {
+		return nil, err
+	}
+
+	result := &apitypes.MachineSegment{}
+	_, err = m.client.Do(ctx, req, result, &reqID, nil)
+	if err != nil {
+		return nil, err
+	}
+
+	return result, nil
 }
--- a/cmd/machines.go
+++ b/cmd/machines.go
@@ -7,6 +7,7 @@ import (
 	"os"
 	"strings"
 	"text/tabwriter"
+	"time"
 
 	"github.com/urfave/cli"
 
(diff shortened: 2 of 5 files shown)


ianlivingstone:
This is missing the `TEAM` or `ROLE` column which would just list the `non-machine-team` membership (since a machine can only belong to one team *other* than the system `machine `team).

You can fetch all the machine teams using the `api.Teams.List` method I introduced for `machines create` and figure out which team defines the role by deriving the orgs `machine` team using the OrgID (see the machine create code I added for how to that).

jeffandersen:
@ianlivingstone added TEAM column, support `--team` flag.

api/machines.go
```diff
@@ -56,3 +57,32 @@ func createTokenSecret() (*base64.Value, error) {
 
 	return base64.NewValue(value), nil
 }
+
+// List machines in the given org and state
+func (m *MachinesClient) List(ctx context.Context, orgIDs []*identity.ID, state *string, teamID *identity.ID) ([]*apitypes.MachineSegment, error) {
```
ianlivingstone: The endpoint only supports a single org id today, so we should probably match that behaviour

cmd/machines.go
```diff
@@ -43,11 +44,101 @@ func init() {
 					setUserEnv, checkRequiredFlags, createMachine,
 				),
 			},
+			{
+				Name:  "list",
+				Usage: "List machines for an organization",
+				Flags: []cli.Flag{
+					orgFlag("org to the machine belongs to", true),
+					teamFlag("the team the machine belongs to", false),
+					destroyedFlag(),
```
ianlivingstone: Usage for destroyed should probably say "Only return destroyed machines" or something along those lines.

cmd/machines.go
```diff
@@ -43,11 +44,101 @@ func init() {
...
+		return errs.NewExitError("Org not found.")
+	}
+	orgID := org.ID
+
+	state := primitive.MachineActiveState
+	if ctx.Bool("destroyed") {
+		state = primitive.MachineDestroyedState
+	}
+
+	teams, err := client.Teams.List(c, org.ID, ctx.String("team"), "")
```
ianlivingstone: We should probably filter by "machine team", so we only return teams that a machine can belong too (via the type property).

cmd/machines.go
```diff
@@ -43,11 +44,101 @@ func init() {
...
+		return errs.NewExitError("Machine teams not found")
+	}
+
+	var teamID *identity.ID
+	if ctx.String("team") != "" {
+		teamID = teams[0].ID
+	}
+
+	orgIDs := []*identity.ID{orgID}
+	machines, err := client.Machines.List(c, orgIDs, &state, teamID)
```
ianlivingstone: You can just pass an empty teamID (`nil`), and it will return all if we haven't found the specified team.

cmd/machines.go
```diff
@@ -43,11 +44,101 @@ func init() {
...
+	if err != nil {
+		return err
+	}
+
+	teamMap := make(map[identity.ID]primitive.Team, len(teams))
+	for _, t := range teams {
+		if t.Body.TeamType == primitive.MachineTeam {
+			teamMap[*t.ID] = *t.Body
```
ianlivingstone: see comment above re: team type parameter to `teams.List`., we wouldn't need this if statement.

cmd/machines.go
```diff
@@ -43,11 +44,101 @@ func init() {
...
+	fmt.Println("")
+	w := tabwriter.NewWriter(os.Stdout, 0, 0, 8, ' ', 0)
+	fmt.Fprintln(w, "ID\tNAME\tSTATE\tTEAM\tCREATION DATE")
+	fmt.Fprintln(w, " \t \t \t \t ")
+	for _, machine := range machines {
+		mID := machine.Machine.ID.String()
+		m := machine.Machine.Body
+		teamName := "-"
+		for _, m := range machine.Memberships {
+			team := teamMap[*m.Body.TeamID]
```
ianlivingstone: we should be checking that the team was found (e.g. `team, ok := teamMap[*m.Body.TeamID]`) if it's not then error out, then we wouldn't need to check the team type again

cmd/machines.go
jeffandersen: `  --destroyed, -d       Filter by destroyed.`

ianlivingstone: By destroyed what though? The other help text explains that this is specific to destroyed machines.

cmd/machines.go
jeffandersen: That's what this is doing unless they're looking up by team.

cmd/machines.go
jeffandersen: The memberships returned include the system team, so we still have to do the if statement

jeffandersen:
@ianlivingstone changes pushed

PR #9, opened 2016-10-31
Machines login

ianlivingstone:
Machines can now login using environment variables passed to the CLI via `TORUS_TOKEN_ID` and `TORUS_TOKEN_SECRET`.

Over time our support for machines will evolve to include managing the daemon as a service.

### DIFF
--- a/api/session.go
+++ b/api/session.go
@@ -154,9 +154,14 @@ func (s *SessionClient) MachineLogin(ctx context.Context, tokenID, tokenSecret s
 		return err
 	}
 
+	secret, err := base64.NewValueFromString(tokenSecret)
+	if err != nil {
+		return err
+	}
+
 	login := apitypes.MachineLogin{
 		TokenID: &ID,
-		Secret:  base64.NewValue([]byte(tokenSecret)),
+		Secret:  secret,
 	}
 
 	rawLogin, err := json.Marshal(login)
--- a/api/session.go
+++ b/api/session.go
@@ -2,44 +2,41 @@ package api
 
 import (
 	"context"
+	"encoding/json"
 	"errors"
 
 	"github.com/manifoldco/torus-cli/apitypes"
+	"github.com/manifoldco/torus-cli/base64"
 	"github.com/manifoldco/torus-cli/envelope"
 	"github.com/manifoldco/torus-cli/identity"
 	"github.com/manifoldco/torus-cli/primitive"
 )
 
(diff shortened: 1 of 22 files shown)


api/session.go
```diff
@@ -2,16 +2,121 @@ package api
...
+	if s.sessionType == apitypes.MachineSession {
+		return "none"
+	}
+
+	return s.identity.Body.(*primitive.User).Email
+}
+
+// NewSession returns a new session constructed from the payload of the current
+// identity as returned from the Daemon
+func NewSession(resp *apitypes.Self) (Session, error) {
```
jbowes: Since `NewSession` is already switching on the type here, how about making `Session` a struct instead of an interface, and pulling the appropriate values to put into the struct out here?

cmd/orgs.go
```diff
@@ -100,16 +100,19 @@ func createOrgByName(c context.Context, ctx *cli.Context, client *api.Client, na
...
 		return err
 	}
 
 	withoutPersonal := orgs
-	for i, o := range orgs {
-		if o.Body.Name == self.Body.Username {
-			fmt.Printf("  %s [personal]\n", o.Body.Name)
-			withoutPersonal = append(orgs[:i], orgs[i+1:]...)
+
+	if session.Type() == apitypes.UserSession {
```
jeffandersen: Should we not be showing some sort of output when a machine tries to list?

jeffandersen: This same change has to be made for `torus ls /` in `orgPaths`

cmd/debug.go
```diff
@@ -109,8 +103,17 @@ func debugInfoCmd(ctx *cli.Context) error {
...
+
+		if session.Type() == apitypes.MachineSession {
+			fmt.Fprintf(w, "%s\t%s\n", "Machine ID", session.ID())
+			fmt.Fprintf(w, "%s\t%s\n", "Machine Token ID", session.AuthID())
+		} else {
+			fmt.Fprintf(w, "%s\t%s\n", "User ID", session.ID())
+			fmt.Fprintf(w, "%s\t%s <%s>\n", "Identity", session.Name(), session.Email())
+		}
+
+		fmt.Fprintf(w, "%s\t%s\n", "Username", session.Username())
```
ianlivingstone: We need a common "concept" because this is the identity slug, so we could just say `Identity`.

PR #19, opened 2016-11-01
Add machines into newly created keyrings

jbowes:
Just like for users, when we need to create a new keyring, ensure that
all machines that should have access to the keyring do in fact have
access.

As we do for users, for now, add *all* machines in the org, and rely on
server-side ACL blocking for access restriction.

### DIFF
--- a/daemon/logic/utils.go
+++ b/daemon/logic/utils.go
@@ -96,12 +96,17 @@ func createCredentialGraph(ctx context.Context, credBody *PlaintextCredential,
 		return nil, err
 	}
 
-	team, err := findMembersTeam(teams)
+	membersTeam, machinesTeam, err := findSystemTeams(teams)
 	if err != nil {
 		return nil, err
 	}
 
-	memberships, err := client.Memberships.List(ctx, credBody.OrgID, team.ID, nil)
+	userMembers, err := client.Memberships.List(ctx, credBody.OrgID, membersTeam.ID, nil)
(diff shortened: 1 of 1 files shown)


daemon/logic/utils.go
```diff
@@ -96,12 +96,17 @@ func createCredentialGraph(ctx context.Context, credBody *PlaintextCredential,
...
 		return nil, err
 	}
 
-	memberships, err := client.Memberships.List(ctx, credBody.OrgID, team.ID, nil)
+	userMembers, err := client.Memberships.List(ctx, credBody.OrgID, membersTeam.ID, nil)
+	if err != nil {
+		return nil, err
+	}
+
+	machineMembers, err := client.Memberships.List(ctx, credBody.OrgID, machinesTeam.ID, nil)
```
ianlivingstone: Right now, destroyed machines are not removed from their teams, I can fix this quickly, but it'll require changes to the way `machines list` and `machines view` work (as they're expecting memberships to come back). 

Thoughts?

ianlivingstone: (The reasoning for leaving them with their teams were to maintain their "role")

jbowes: their token is destroyed though, so finding the subject for encryption key selection will filter it out :)

ianlivingstone: I totally misread the code 👍

PR #33, opened 2016-11-07
Allow user to update their name or email

jeffandersen:
- `torus profile update`
- Depends on a new route in the registry
- Allows user to update name or email address
  - Email will require re-verification on change

### DIFF
--- a/cmd/profile.go
+++ b/cmd/profile.go
@@ -53,14 +53,18 @@ func profileView(ctx *cli.Context) error {
 	if err != nil {
 		return errs.NewErrorExitError("Error fetching user details", err)
 	}
-	if session.Type() == apitypes.MachineSession {
-		return errs.NewExitError("Machines do not have profiles")
-	}
 
 	w := tabwriter.NewWriter(os.Stdout, 2, 0, 1, ' ', 0)
-	fmt.Fprintf(w, "Name:\t%s\n", session.Name())
-	fmt.Fprintf(w, "Email:\t%s\n", session.Email())
-	fmt.Fprintf(w, "Username:\t%s\n\n", session.Username())
+	if session.Type() == apitypes.MachineSession {
+		fmt.Fprintf(w, "Machine ID:\t%s\n", session.ID())
+		fmt.Fprintf(w, "Machine Token ID:\t%s\n", session.AuthID())
+		fmt.Fprintf(w, "Machine Name:\t%s\n\n", session.Username())
+	} else {
+		fmt.Fprintf(w, "Name:\t%s\n", session.Name())
+		fmt.Fprintf(w, "Email:\t%s\n", session.Email())
+		fmt.Fprintf(w, "Username:\t%s\n\n", session.Username())
+	}
+
 	w.Flush()
 
 	return nil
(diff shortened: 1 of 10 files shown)


jeffandersen:
Having an issue where the local session is cache and calling Set on db is not updating it https://github.com/manifoldco/torus-cli/blob/update-email/daemon/logic/session.go#L63

Digging into that before this is rtg.

daemon/logic/session.go
```diff
@@ -52,6 +52,17 @@ func (s *Session) Login(ctx context.Context, creds apitypes.LoginCredential) err
 	return s.engine.session.Set(self.Type, self.Identity, self.Auth, creds.Passphrase(), authToken)
 }
 
+// Refresh retrieves the user's identity and updates the stored copy of the session
```
ianlivingstone: Why not just have the daemon handle the username/email/password changes instead of proxying?

The "daemon cache" and it's responsibility should be transparent to the CLI.

ianlivingstone: e.g. `PATCH /users/self` or similar on the daemon that also results in update the cache and returns the update `self` to the callee.

cmd/profile.go
```diff
@@ -0,0 +1,117 @@
...
+	"github.com/manifoldco/torus-cli/apitypes"
+	"github.com/manifoldco/torus-cli/config"
+	"github.com/manifoldco/torus-cli/errs"
+
+	"github.com/urfave/cli"
+)
+
+func init() {
+	profile := cli.Command{
+		Name:     "profile",
```
ianlivingstone: We should probably also introduce a reciprocal `profile view` command that lists the email, full name, etc (only accessible by users).

We'd then make `torus status` only about the current "context" you are operating in and nothing more.

cmd/profile.go
```diff
@@ -0,0 +1,117 @@
...
+		return err
+	}
+
+	warning := "\nYou are about to update your profile to the values above."
+	if email != ogEmail {
+		warning = "\nYou will be required to re-verify your email address before taking any further actions within Torus."
+	}
+
+	if ogEmail == email && ogName == name {
+		fmt.Println("\nNo changes made :)")
```
ianlivingstone: 👊

cmd/profile.go
jbowes: 💰

cmd/profile.go
```diff
@@ -0,0 +1,141 @@
...
+		Name:     "profile",
+		Usage:    "Manage your Torus account",
+		Category: "ACCOUNT",
+		Subcommands: []cli.Command{
+			{
+				Name:  "view",
+				Usage: "View your profile",
+				Action: chain(
+					ensureDaemon, ensureSession, loadDirPrefs, loadPrefDefaults,
+					setUserEnv, profileView,
```
jbowes: you only need `ensureDaemon, ensureSession, <func>` for these two.

cmd/profile.go
```diff
@@ -0,0 +1,141 @@
...
+
+	client := api.NewClient(cfg)
+	c := context.Background()
+
+	session, err := client.Session.Who(c)
+	if err != nil {
+		return errs.NewErrorExitError("Error fetching user details", err)
+	}
+	if session.Type() == apitypes.MachineSession {
+		return errs.NewExitError("Machines do not have profiles")
```
jbowes: :( poor machines

jeffandersen: k they can view them now that we moved identity out of status

cmd/profile.go
jeffandersen: reduced

daemon/logic/session.go
ianlivingstone: This isn't used anywhere now!

jeffandersen: yeah it's gone

PR #34, opened 2016-11-07
Expose Machine Teams as Roles

ianlivingstone:
This pull request introduces the concept of "machine roles" to the CLI which are powered using the "team" primitive.

The fundamental difference between a user and a machine is that a machine can only belong to *one* team other than the main `machines` system team which is the team that defines the role of the machine.

Included in this request:

- Bug fix to prevent a panic occurring when listing teams with an unknown org
- Remove capability to list machines through `torus teams list` or `torus teams members`
- Introduce the `torus machines roles` and `torus machines create-role` commands for listing and creating machine roles.

**UI**

```
Ians-MacBook-Pro-2:torus-cli ianlivingstone$ ./torus machines roles
machine             [system]
hellomachines
joe
axe
fdf

All machines belong to the "machine" role.
Ians-MacBook-Pro-2:torus-cli ianlivingstone$ ./torus machines create-role api
✔ Org name: ian
✔ Role name: api

Role api created.
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,16 @@ _Unpublished_
 - Introduced `--user, -u` and `--machine, -m` flags to `torus set`, `torus
   unset`, `torus view`, `torus run`, and `torus ls` for specifying machine or
   user identity
+- Introduce `machines roles list` and `machines roles create` commands for
+  viewing and creating machine roles.
+- Machine teams no longer appear under `teams list` nor can you view machine
+  teams through `teams members`.
+- The `machines` command now appears under the `ORGANIZATIONS` category when
+  listing commands with `torus help`.
+
+**Fixes**
+
+- Listing teams no longer results in a panic when an unknown org is specified.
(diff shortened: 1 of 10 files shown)


ianlivingstone:
I can squash these down into two commits once reviewed, I just wanted to make sure the changes were easy to digest.

jbowes:
`torus machines roles list` && `torus machines roles create` would match the existing commands better.

primitive/primitive.go
```diff
@@ -563,6 +563,14 @@ const (
 	MachineTeam = "machine"
 )
 
+// Teams are used to represent a group of identities and their associated
```
jbowes: associated... ?

CHANGELOG.md
```diff
@@ -2,13 +2,25 @@
 
 ## UNPUBLISHED
 
+<<<<<<< HEAD
```
jbowes: oops

PR #36, opened 2016-11-08
Support --user and --machine flags for set, etc.

ianlivingstone:
Introduce flags for specifying user or machine identities through
commands that manipulate secrets (set, unset, run, view, etc).

The machine flag automatically prepends `machine-`, ensuring users don't
have to worry about the complexities of the credential path.

Stringing together multiple `--user` and `--machine` flags will turn
them into an OR expression (e.g. `[ian|machine-api-staging-1231fd]`).

### DIFF
--- a/cmd/set.go
+++ b/cmd/set.go
@@ -22,8 +22,8 @@ var setUnsetFlags = []cli.Flag{
 		"", "TORUS_ENVIRONMENT", true),
 	newSlicePlaceholder("service, s", "SERVICE", "Use this service.",
 		"default", "TORUS_SERVICE", true),
-	newSlicePlaceholder("user, u", "USER", "Use this user (identity).",
-		"*", "TORUS_USER", true),
+	newSlicePlaceholder("user, u", "USER", "Use this user.", "*", "TORUS_USER", false),
+	newSlicePlaceholder("machine, m", "MACHINE", "Use this machine.", "*", "TORUS_MACHINE", false),
 	newSlicePlaceholder("instance, i", "INSTANCE", "Use this instance.",
 		"*", "TORUS_INSTANCE", true),
 }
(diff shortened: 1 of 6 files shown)


jbowes:
Would it be clearer to have the `machine` flag be an alias of `user`, so they'd show up on the same line in the help display?

ianlivingstone:
> Would it be clearer to have the machine flag be an alias of user, so they'd show up on the same line in the help display?

We actually need to differentiate between `--user` and `--machine` internally, if they're an alias I don't think we can detect the differentiation.

jbowes:
> We actually need to differentiate between --user and --machine internally, if they're an alias I don't think we can detect the differentiation.

👍

cmd/set.go
```diff
@@ -22,8 +22,8 @@ var setUnsetFlags = []cli.Flag{
 		"", "TORUS_ENVIRONMENT", true),
 	newSlicePlaceholder("service, s", "SERVICE", "Use this service.",
 		"default", "TORUS_SERVICE", true),
-	newSlicePlaceholder("user, u", "USER", "Use this user (identity).",
-		"*", "TORUS_USER", true),
+	newSlicePlaceholder("user, u", "USER", "Use this user.", "", "TORUS_USER", false),
+	newSlicePlaceholder("machine, m", "MACHINE", "Use this machine.", "", "TORUS_MACHINE", false),
```
jbowes: set no defaults to an empty identity, which is invalid :)
you need to bring back that asterisk

cmd/view.go
```diff
@@ -67,6 +69,80 @@ func viewCmd(ctx *cli.Context) error {
...
+	if strings.Contains(identity, "machine-") {
+		return "", errs.NewExitError(
+			"Do not prepend 'machine-' when using --machine")
+	}
+
+	return "machine-" + identity, nil
+}
+
+// Derives a slice of identity segments for us in building a PathExp object.
+func deriveIdentitySlice(ctx *cli.Context) ([]string, error) {
```
jbowes: These funcs should be in `flags.go` imo

cmd/set.go
ianlivingstone: If you look at deriveIdentity it falls back to the user or machines name!

PR #38, opened 2016-11-08
Extend email in use error to include username

jeffandersen:
We should surface this context until we can update the server to provide more granularity.


jbowes:
LGTM

PR #50, opened 2016-11-10
Add deb/apt releases

jbowes:
The package is all the same, but releases are broken out for the most
recent debian release, and the last two ubuntu releases. This lets us
craft an apt sources.list line that is familiar to users (including the
distro code name), and provides a nice cut-point for supporting upstart
vs systemd, if we can't fit them both into a single deb.


ianlivingstone:
This looks good, when/where will we update/advertise how to acquire these packages?

PR #52, opened 2016-11-13
Generate static directory listings for get.torus.sh

jbowes:
tools/cdn-indexer reads the existing files from s3 and generates a
templated index file, then writes it back to s3. It is run during the
`release-all` Makefile target.

This is on top of #50. I'll rebase after that's merged.

### DIFF
--- a/tools/cdn-indexer/main.go
+++ b/tools/cdn-indexer/main.go
@@ -1,0 +1,165 @@
+package main
+
+import (
+	"bytes"
+	"flag"
+	"log"
+	"path/filepath"
+	"sort"
+	"strings"
+	"text/template"
+	"time"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/session"
+	"github.com/aws/aws-sdk-go/service/s3"
+)
+
+var t *template.Template
+
+var (
+	bucket    string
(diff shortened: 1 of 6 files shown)


tools/cdn-indexer/main.go
```diff
@@ -0,0 +1,168 @@
...
+			Body:         bytes.NewReader(buf.Bytes()),
+			ContentType:  aws.String("text/html"),
+			CacheControl: aws.String("public, max-age=300"),
+		}
+		_, err := svc.PutObject(params)
+		if err != nil {
+			log.Fatal(err)
+		}
+
+		//os.MkdirAll(dirPath, 0755)
```
ianlivingstone: several commented out lines?

ianlivingstone:
Also, can we set the content-type for the shasum file ?

jbowes:
> Also, can we set the content-type for the shasum file ?

That should be done during upload of those files. They should also have explicit cache headers. I'd rather handle both together, and in another PR.

ianlivingstone:
> That should be done during upload of those files. They should also have explicit cache headers. I'd rather handle both together, and in another PR.

Sounds good, as long as we track them!

jbowes:
https://github.com/manifoldco/torus-cli/issues/55

PR #59, opened 2016-11-14
Allow bypassing torus root permission checks

jbowes:
For the system wide daemon machine case, we want to have looser
permissions on the torus root dir.

Introduce a hidden flag to bypass permission checks on the torus root
dir, so we can allow the system install to take care of permissions
instead.

### DIFF
--- a/cmd/daemon.go
+++ b/cmd/daemon.go
@@ -48,6 +48,11 @@ func init() {
 						Usage:  "Run as a background session daemon",
 						Hidden: true, // not displayed in help; used internally
 					},
+					cli.BoolFlag{
+						Name:   "no-permission-check",
+						Usage:  "Skip Torus root dir permission checks",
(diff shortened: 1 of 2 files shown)


ianlivingstone:
What's the reasoning for hiding it?

jbowes:
> What's the reasoning for hiding the flag?

I don't think it's the idea long term solution.
when we are enforcing user or group id on the socket messages themselves, it won't be the right method.

It'll probably have to switch to a value in the torusrc, which will become an /etc/torus.conf...

So, I'm not completely sold on it. It does the job, works fine, but isn't future proof.

ianlivingstone:
Okay makes sense, perhaps we should open a discussion issue?

jbowes:
> Okay makes sense, perhaps we should open a discussion issue?

Yeah -> #60

PR #62, opened 2016-11-14
Include a systemd service with our rpm

jbowes:
The service can be configured with machine token credentials in
/etc/torus/token.environment, for use in production environments.

Other processes running on the system can access secrets from torus by:
- running as a user in the `torus` group
- setting the TORUS_ROOT environment variable to /var/run/torus

While there is a torus user, this user should *not* be used for running
processes that need access to secrets. It is an isolated sandbox that
has access to the machine's token credentials.

### DIFF
--- a/contrib/systemd/torus.service
+++ b/contrib/systemd/torus.service
@@ -1,0 +1,14 @@
+[Unit]
+Description=Torus daemon
+Documentation=https://www.torus.sh/docs/
+After=network.target
+
+[Service]
+Type=simple
+User=torus
+Environment="TORUS_ROOT=/var/run/torus"
+EnvironmentFile=/etc/torus/token.environment
+ExecStart=/usr/bin/torus daemon start --foreground --no-permission-check
+
+[Install]
+WantedBy=multi-user.target
--- a/daemon/socket/proxy.go
+++ b/daemon/socket/proxy.go
@@ -48,6 +48,11 @@ type AuthProxy struct {
 
 // NewAuthProxy returns a new AuthProxy. It will return an error if creation
 // of the domain socket fails, or the upstream registry URL is misconfigured.
+//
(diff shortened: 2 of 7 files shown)


daemon/socket/proxy.go
```diff
@@ -48,10 +48,10 @@ type AuthProxy struct {
 
 // NewAuthProxy returns a new AuthProxy. It will return an error if creation
 // of the domain socket fails, or the upstream registry URL is misconfigured.
-func NewAuthProxy(c *config.Config, sess session.Session, db *db.DB,
```
ianlivingstone: Can we get a doc string about the different the socket is configured (read/write user/group)?

contrib/systemd/torus.service
```diff
@@ -0,0 +1,14 @@
...
+Description=Torus daemon
+Documentation=https://www.torus.sh/docs/
+After=network.target
+
+[Service]
+Type=simple
+User=torus
+Environment="TORUS_ROOT=/var/run/torus"
+EnvironmentFile=/etc/torus/token.environment
+ExecStart=/usr/bin/torus daemon start --foreground --no-permission-check
```
ianlivingstone: So in "service mode" anyone on the system gets access to any of the secrets that the systems identity provides it?

ianlivingstone: I see -- it's constrained to the group. Does this also mean that they can access the TOKEN_ID and TOKEN_SECRET via `/proc` for the running user? Or is the idea that the user is locked down, then you just share secrets via group?

daemon/socket/proxy.go
jbowes: *nods*

contrib/systemd/torus.service
ianlivingstone: 👍 Awesome.

PR #63, opened 2016-11-15
Update our error strings for new the golint check

jbowes:


ianlivingstone:
Is this something we should standardize throughout? We'll now have mismatches on capitalization.

jbowes:
I believe the intention with this lint check is to keep your error strings short and not at all attractive, so they aren't what you end up showing to users.

Your presentation layer should take care of making things handsome for users, and your library should provide a richer interface than just the vanilla error type, if needed: https://golang.org/doc/effective_go.html#errors https://github.com/golang/go/wiki/CodeReviewComments#error-strings

We don't do this for some of our errors, but i think we do for many of them already!

jbowes:
what's annoying is the change to golint to support detecting newlines in error messages changed the confidence level for messages starting with caps, meaning they now trigger for us, as a side effect.

it's a good change overall, but not a nice surprise.

PR #64, opened 2016-11-15
Fix build times with gosimple

jbowes:
Use it via gometalinter now, so imports are all correct.

Fix the identified issues, and use a temporary fork of go-bindata so it
won't generate un-simple files.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -10,8 +10,13 @@ TARGETS=\
 
 VERSION?=$(shell git describe --tags --abbrev=0 | sed 's/^v//')
 
+LINTERS=\
+	gosimple \
+	misspell \
+	ineffassign
+
 all: binary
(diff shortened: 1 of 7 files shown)


Makefile
```diff
@@ -10,20 +10,27 @@ TARGETS=\
...
 #################################################
 
+# XXX: switch back to upstream go-bindata when it passes gosimple
+
 BOOTSTRAP=\
 	github.com/Masterminds/glide \
 	github.com/golang/lint/golint \
 	honnef.co/go/simple/cmd/gosimple \
-	github.com/jteeuwen/go-bindata/... \
+	github.com/jbowes/go-bindata/... \
```
jeffandersen: Should probably just move this to Manifold

jbowes: its in manifold now

PR #72, opened 2016-11-17
Output release artifact locations after release-all

### LINKED ISSUE #69: Output release url after release-all step
Once a user has run the `release-all` build step, we don't actually know where the generated artifacts live.

It'd be good for the `makefile` to produce that url. Secondly, the "staging" or "release" bucket used for non-production releases should be indexable so we can browse our release candidates.

This should be publicly available.

jbowes:
Everything is output in a nice format, for easy pasting around.

Closes #69

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -177,11 +177,9 @@ deb-container:
 
 RELEASE_ENV?=stage
 ifeq (stage,$(RELEASE_ENV))
-	TORUS_S3_BUCKET=s3://releases.arigato.sh
-	BOTTLE_URL=https://s3.amazonaws.com/releases.arigato.sh
+	TORUS_S3_BUCKET=prerelease.torus.sh
 else ifeq (prod,$(RELEASE_ENV))
-	TORUS_S3_BUCKET=s3://get.torus.sh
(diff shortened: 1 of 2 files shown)


jbowes:
Looks like this:
```
==================================================================================
Release 0.17.0 is ready for stage

zips: https://prerelease.torus.sh/0.17.0/

npm:  npm install -g https://prerelease.torus.sh/npm/0.17.0/torus.tar.gz
brew: brew install https://prerelease.torus.sh/brew/0.17.0/torus.rb

yum:
sudo tee /etc/yum.repos.d/torus.repo <<-'EOF'
[torus]
name=torus-cli repository
baseurl=https://prerelease.torus.sh/rpm//
enabled=1
gpgcheck=0
EOF
sudo yum install torus # or dnf

deb:
DISTRO=$(lsb_release -i | awk '{print tolower($3)}')
CODENAME=$(lsb_release -c | awk ‘{print $2})’
sudo tee /etc/apt/sources.list.d/torus.list \
    <<< "deb https://prerelease.torus.sh/$DISTRO/ $CODENAME main"
sudo apt-get update
sudo apt-get install torus

==================================================================================
```

Issue #51, opened 2016-11-11
template command - exporting to a file


ianlivingstone:
## Problem

Configuration files are still core to many deployment flows and many developers prefer them.

Our challenge is to offer an experience that enables you to store secrets with torus while integrating seamlessly into your pre-existing deployment pipeline.

To do that, we can offer people the ability to derive configuration files at deployment time accepting the fact that these files must be kept "safe" (as they move outside the confines of the Torus ecosystem).

## Proposal

We're proposing a `torus template` command that operates exactly like `torus run` and `torus view` allow users to specify the service, environment, etc., through flags but also require a template file (e.g. `torus template -e production -s api config.json.template`).

Torus would then use the given template to generate a config file and print it to stdout. Optionally, if `--file` is provided then we will write the output to the given file path. 

For simplicity and approachability, I think the best option for a templating engine would be something like [handlebars](http://handlebarsjs.com/) which has a [decent go library](https://github.com/aymerick/raymond).

If an expected variable is not present, the template command will fail.

#### Command Help

```
NAME:
   torus template - Generate a configuration file using secrets for the given project, service, and environment. By default, the result is written to stdout.

USAGE:
   torus template [template-file] [output file] [command options]

CATEGORY:
   SECRETS

OPTIONS:
   --org ORG, -o ORG                 Use this organization.
   --project PROJECT, -p PROJECT     Use this project.
   --environment ENV, -e ENV         Use this environment.
   --service SERVICE, -s SERVICE     Use this service. (default: default)
   --user USER, -u USER              Use this user.
   --machine MACHINE, -m MACHINE     Use this machine.
   --instance INSTANCE, -i INSTANCE  Use this instance. (default: 1)
```

#### Example Usage

```
$ cat service.json.template
{
  "port": {{ port }},
  "sendgrid_token": {{ sendgrid_token }}
}

$ torus template -e production -s api service.json.template
{
  "port": 3211,
  "sendgrid_token": "wwerwerwerwrwrwre"
}

$ torus template -e production -s auth service.json.template auth.json

# For complex config files, we can offer an iterable (config) for looping over the available configuration.
$ cat complex.env.template
{
  "port": {{ config.port }},
  "sendgrid_token": "{{ config.sendgrid_token }}"
}
```

## Todo

- [x] Implement new `--format` flag with `torus view`
- [ ] Implement `torus template` command as spec'd above

ianlivingstone:
[...]

PR #78, opened 2016-11-18
Introduce torus template command

ianlivingstone:
The `torus template` command allows a user to inject secrets stored in
torus as they derive a configuration file from a template.

The current templating engine is the [raymond
handlebars](https://github.com/aymerick/raymond) engine for Go
which supports v3 of the handlebars spec.

### Example Usage

Simple variable replacement example with a json file:

```bash
$ cat test.json.template
{
    "message": "{{ message }}",
    "port": {{ port }}
}
$ ./torus template test.json.template
{
    "message": "this value was set using ag",
    "port": 34234
}
```

Outputting to a file instead of stdout using the optional output parameter:

```bash
$ ./torus template test.json.template config.json
$ cat config.json
{
    "message": "this value was set using ag",
    "port": 34234
}
```

Iterating over the special `_config` variable:

```bash
$ cat test.json.template
{
    "message": "{{ message }}",
    "port": {{ port }},
    "config": {
    {{#each _config}}
        "{{ name }}": "{{ value }}"{{#unless @last}},{{/unless}}
    {{/each}}
    }
}
$ ./torus template test.json.template
{
    "message": "this value was set using ag",
    "port": 4432,
    "config": {
        "message": "this value was set using ag",
        "port": "4432",
        "test_db": "testdburl"
    }
}
```

Closes manifoldco/torus-cli#51

### DIFF
--- a/promptui/select.go
+++ b/promptui/select.go
@@ -25,7 +25,7 @@ func (s *Select) Run() (int, string, error) {
 }
 
 func (s *Select) innerRun(starting int, top rune) (int, string, error) {
-	stdin := readline.NewCancelableStdin()
+	stdin := readline.NewCancelableStdin(os.Stdin)
 	c := &readline.Config{}
 	err := c.Init()
 	if err != nil {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # CHANGELOG
 
 ## v0.18.0
+
 _Unreleased_
 
 **Notable Changes**
@@ -11,6 +12,10 @@ _Unreleased_
   group can access it. To run the unit, both `TORUS_TOKEN_ID` and
   `TORUS_TOKEN_SECRE` must be set in `/etc/torus/token.environment`. See v0.17.0
   for the matching rpm change.
+- Introduced `torus template [flags] <template-file> [output-file]` for
+  generating a configuration file from a template. The underlying templating
+  engine is [handlebars](https://github.com/aymerick/raymond) allowing Torus to
+  be used in any file based configuration flow.
 
 **Fixes**
 
--- a/cmd/template.go
+++ b/cmd/template.go
@@ -1,13 +1,15 @@
 package cmd
 
 import (
+	"bytes"
 	"fmt"
 	"io/ioutil"
 	"os"
 	"path/filepath"
+	"reflect"
 	"strings"
+	"text/template"
 
-	"github.com/aymerick/raymond"
 	"github.com/urfave/cli"
 
 	"github.com/manifoldco/torus-cli/apitypes"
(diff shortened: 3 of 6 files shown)


ianlivingstone:
This works great for the most basic use-case, however, to iterate over the exposed configuration items and inject them into a json file isn't exactly ideal:

```
$ cat test.json.template
{
    "message": "{{ message }}",
    "port": {{ port }},
    "config": {
    {{#each _config}}
        "{{ name }}": "{{ value }}"{{#unless @last}},{{/unless}}
    {{/each}}
    }
}
$ ./torus template test.json.template
{
    "message": "this value was set using ag",
    "port": 4432,
    "config": {
        "message": "this value was set using ag",
        "port": "4432",
        "test_db": "testdburl"
    }
}
```

A few problems:

- Your template has to produce valid json -- so you'll have to do that yourself
- We don't offer a good helper to do the type outputting properly
- We don't offer a helper for converting an entire slice or map to JSON inside the template itself

These will be problems despite the engine we use for templating -- as the engine will never be aware enough to offer nice solutions for deriving JSON, YAML, or <insert config here>. To make this a nice experience, we'd have to offer helpers.

I'm not super happy with the more advanced use cases, however, it seems to me that as you got more advanced you'd probably be using something like ansible or chef to derive your configuration files. At which point, you wouldn't want to shell out, you'd want a native plugin.

Thoughts?

ianlivingstone:
> Thoughts?

For the use cases this feature supports (I have a config file for a service, I want to put secrets inside it) the feature as coded works well. The challenges I noted above, only comes into effect if you are storing complex data (which we don't support today) *or* you want to combine multiple service configs into one config file (which this feature does not support).

Therefore, for the purposes of enabling this workflow, I think we can go ahead and merge this as implemented. Then we can begin discussing how we'd want to enable those other use cases (which I'd advocate offering language libraries and plugins to things like ansible, chef or puppet).

promptui/select.go
```diff
@@ -25,7 +25,7 @@ func (s *Select) Run() (int, string, error) {
 }
 
 func (s *Select) innerRun(starting int, top rune) (int, string, error) {
-	stdin := readline.NewCancelableStdin()
+	stdin := readline.NewCancelableStdin(os.Stdin)
```
jeffandersen: Is this due to a change in the underlying package?

Should probably be noted as a dep upgrade/fix in the changelog if we're changing more than what goes into the feature.

ianlivingstone: Yep -- the dep got upgraded.

Should we document dependency upgrades? It doesn't seem useful to anyone but those who contribute to the project.

jeffandersen:
Values that are found in the template but not actually set in the environment come out as `""`.

Should they not be `null`?

jeffandersen:
I guess the problem there is that they're likely wrapped in quotes in the template...

ianlivingstone:
Yeah -- ideally we'd error but I'm not sure if the template engine we're using provides a hook for catching non-provided template variables which would be ideal.

ianlivingstone:
Going to look for a better package before we move ahead with this change.

PR #88, opened 2016-11-25
Support creating missing keypairs via worklog

jbowes:
```
$ ./torus worklog list --org badkeys
IDENTITY         TYPE      SUBJECT
0a74q6462ymxf5g  keypairs  badkeys
$ ./torus worklog resolve --org badkeys
✔  0a74q6462ymxf5g  Keypairs generated.
$ ./torus worklog list --org badkeys
IDENTITY  TYPE  SUBJECT
```

### DIFF
--- a/daemon/logic/utils.go
+++ b/daemon/logic/utils.go
@@ -368,19 +368,23 @@ func createKeyringMemberships(ctx context.Context, c *crypto.Engine, client *reg
 	return v1members, v2members, nil
 }
 
-// fetchKeyPairs fetches the user's signing and encryption keypairs from the
-// registry for the given org id.
-func fetchKeyPairs(ctx context.Context, client *registry.Client,
-	orgID *identity.ID) (*identity.ID, *identity.ID, *crypto.KeyPairs, error) {
+// fetchRegistryKeyPairs fetches the user's signing and encryption keypairs
+// from the registry for the given org id.
+// It returns an error if the keypairs cannot be fetched from the registry,
+// or if an invalid keypair type is seen.
+// It returns nil for keypair types that are not found.
+func fetchRegistryKeyPairs(ctx context.Context, client *registry.Client,
+	orgID *identity.ID) (*registry.ClaimedKeyPair, *registry.ClaimedKeyPair, error) {
 
 	keyPairs, err := client.KeyPairs.List(ctx, orgID)
 	if err != nil {
-		return nil, nil, nil, err
+		return nil, nil, err
 	}
 
-	var sigClaimed registry.ClaimedKeyPair
-	var encClaimed registry.ClaimedKeyPair
-	for _, keyPair := range keyPairs {
+	var sigClaimed *registry.ClaimedKeyPair
+	var encClaimed *registry.ClaimedKeyPair
+	for _, kp := range keyPairs {
+		var keyPair = kp
 		if keyPair.Revoked() {
 			continue
 		}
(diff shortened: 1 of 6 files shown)


daemon/logic/worklog.go
```diff
@@ -70,6 +72,21 @@ func (w *Worklog) List(ctx context.Context, orgID *identity.ID) ([]apitypes.Work
 		items = append(items, item)
 	}
 
+	keypairs, err := w.engine.client.KeyPairs.List(ctx, org.ID)
+	if err != nil {
+		return nil, err
+	}
+
+	if len(keypairs) == 0 {
```
ianlivingstone: What about revoked keypairs? technically those are missing as well -- if a user has no keys for an org.

jbowes: troo

daemon/logic/utils.go
```diff
@@ -368,38 +368,55 @@ func createKeyringMemberships(ctx context.Context, c *crypto.Engine, client *reg
...
-// fetchKeyPairs fetches the user's signing and encryption keypairs from the
-// registry for the given org id.
-func fetchKeyPairs(ctx context.Context, client *registry.Client,
-	orgID *identity.ID) (*identity.ID, *identity.ID, *crypto.KeyPairs, error) {
+// fetchRegistryKeyPairs fetches the user's signing and encryption keypairs
+// from the registry for the given org id.
+// It returns an error if the keypairs cannot be fetched from the registry,
+// or if an invalid keypair type is seen.
+// It returns nil for keypair types that are not found.
+func fetchRegistryKeyPairs(ctx context.Context, client *registry.Client,
```
ianlivingstone: 👍

ianlivingstone: This assumes a user only ever has at most one keypair for an organization.

Is there any situation where they might have two?

daemon/logic/worklog.go
```diff
@@ -70,6 +72,21 @@ func (w *Worklog) List(ctx context.Context, orgID *identity.ID) ([]apitypes.Work
...
 
+	encClaimed, sigClaimed, err := fetchRegistryKeyPairs(ctx, w.engine.client, org.ID)
+	if err != nil {
+		return nil, err
+	}
+
+	if encClaimed == nil || sigClaimed == nil {
+		item := apitypes.WorklogItem{
+			Subject: org.Body.(*primitive.Org).Name,
+			Summary: "Keypair missing for org.",
```
ianlivingstone: perhaps: `Signing or encryption keypair missing for org.` ?

daemon/logic/worklog.go
```diff
@@ -112,7 +129,21 @@ func (w *Worklog) Resolve(ctx context.Context, orgID *identity.ID,
 			State:   apitypes.ManualWorklogResult,
 			Message: "Please set a new value for the secret at " + item.Subject,
 		}, nil
+	case apitypes.MissingKeypairsWorklogType:
+		err = w.engine.GenerateKeypair(ctx, n, orgID)
+		if err != nil {
```
ianlivingstone: This will end up hiding the actual underlying error - how do we get at that for support / provide information to our users (e.g. their session expired so they need to login)

daemon/logic/utils.go
jbowes: not at present, afaik.

daemon/logic/worklog.go
jbowes: heck i can give a different message for all three cases :)

daemon/logic/worklog.go
jbowes: i'll include the error

ianlivingstone: :+1

PR #90, opened 2016-11-28
Add prefs for core.hints and core.progress

jeffandersen:
- Introduces `ui` package, which will be home to any ui output functions
  - Will come in handy as we standardize all outputs per the style guide
- Adds prefs for `core.hints` and `core.progress` to enable (default) or disable these outputs

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,14 @@
 # CHANGELOG
 
+## v0.19.0
+
+_Unreleased_
+
+**Notable Changes**
+
+- Add preferences for `core.disable_progress` and `core.disable_hints` to control levels of output
+  in preparation for additional onboarding output.
+
 ## v0.18.0
 
 _2016-11-30_
--- a/main.go
+++ b/main.go
@@ -16,7 +16,7 @@ func main() {
 		cmd.VersionLookup(ctx)
 	}
 
-	preferences, _ := prefs.NewPreferences(true)
+	preferences, _ := prefs.NewPreferences()
 	ui.Init(preferences)
 
 	app := cli.NewApp()
--- a/main.go
+++ b/main.go
@@ -7,12 +7,17 @@ import (
 
 	"github.com/manifoldco/torus-cli/cmd"
 	"github.com/manifoldco/torus-cli/config"
+	"github.com/manifoldco/torus-cli/prefs"
+	"github.com/manifoldco/torus-cli/ui"
 )
 
 func main() {
 	cli.VersionPrinter = func(ctx *cli.Context) {
 		cmd.VersionLookup(ctx)
 	}
+
+	preferences, _ := prefs.NewPreferences(true)
+	ui.Init(preferences)
 
 	app := cli.NewApp()
 	app.Version = config.Version
--- a/ui/ui.go
+++ b/ui/ui.go
@@ -1,0 +1,35 @@
+package ui
+
+import (
+	"fmt"
+
+	"github.com/manifoldco/torus-cli/prefs"
+)
+
+// enableProgress is whether progress events should be displayed
+var enableProgress = false
+
+// enableHints is whether hints should be displayed
+var enableHints = false
+
+// Init prepares the ui preferences
+func Init(preferences *prefs.Preferences) {
+	enableProgress = preferences.Core.EnableProgress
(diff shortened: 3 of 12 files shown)


ui/ui.go
```diff
@@ -0,0 +1,48 @@
+package display
```
jbowes: :D

ui/ui.go
```diff
@@ -0,0 +1,48 @@
...
+	"github.com/manifoldco/torus-cli/prefs"
+)
+
+const (
+	InfoLevel  = 3
+	ErrorLevel = 1
+)
+
+// LogLevel represents the current allowed level of output
+var LogLevel int
```
jbowes: The loglevel stuff isn't (and IMO shouldn't be) used

ui/ui.go
```diff
@@ -0,0 +1,48 @@
...
+
+// EnableProgress is whether progress events should be displayed
+var EnableProgress = true
+
+// EnableHints is whether hints should be displayed
+var EnableHints = true
```
jbowes: These vars shouldn't be public, to discourage people fiddling with them directly

ui/ui.go
```diff
@@ -0,0 +1,48 @@
...
+
+func init() {
+	preferences, _ := prefs.NewPreferences(true)
```
jbowes: maybe the prefs should get passed in via another func?

ianlivingstone: 👍

jeffandersen: You're saying a custom init function?

Cause it would definitely suck to pass prefs every time you were doing an output.

ui/ui.go
```diff
@@ -0,0 +1,40 @@
...
+var disableHints = false
+
+// Init prepares the ui preferences
+func Init(preferences *prefs.Preferences) {
+	if preferences.Core.DisableProgress {
+		disableProgress = true
+	}
+	if preferences.Core.DisableHints {
+		disableHints = true
+	}
```
jbowes:
```go
disableProgress = preferences.Core.DisableProgress
disableHints = preferences.Core.DisableHints
```

jeffandersen:
I've changed preferences to display default values with `torus prefs list` (and save them, when using `torus prefs set`).

```
> torus prefs list

/Users/jeffandersen/.torusrc

[core]
    registry_uri = https://registry.arigato.sh
    context      = true
    progress     = true
    hints        = true

```

ui/ui.go
```diff
@@ -0,0 +1,35 @@
...
+
+// Init prepares the ui preferences
+func Init(preferences *prefs.Preferences) {
+	enableProgress = preferences.Core.EnableProgress
+	enableHints = preferences.Core.EnableHints
+}
+
+// Progress handles the ui output for progress events, when enabled
+func Progress(str string) {
+	if enableProgress {
```
jbowes: `!enableProgress` (likewise for hints).

CHANGELOG.md
```diff
@@ -16,6 +16,8 @@ _2016-11-30_
 - Teach `worklog` about approving invites.
 - Unhide `worklog resolve`, as it can now be used to generate missing keypairs
   for an org, or approve an invite.
+- Add preferences for `core.disable_progress` and `core.disable_hints` to control levels of output
+  in preparation for additional onboarding output.
```
jbowes: This changelog entry has to be in a new release now

main.go
```diff
@@ -7,13 +7,18 @@ import (
...
+	"github.com/manifoldco/torus-cli/prefs"
+	"github.com/manifoldco/torus-cli/ui"
 )
 
 func main() {
 	cli.VersionPrinter = func(ctx *cli.Context) {
 		cmd.VersionLookup(ctx)
 	}
 
+	preferences, _ := prefs.NewPreferences(true)
```
jbowes: NewPreferences is now only called with `true`. Mayaswell drop the bool arg.

jbowes: This goes back to displaying *all* prefs with a prefs list, and setting defaults into the file on a prefs set.

I guess that's better than having to do the negated option values though

jeffandersen: yeah this was the thing I was talking to you about, having the default booleans appear in the rc file after save.

PR #95, opened 2016-11-29
Remove duplicate ? for change password

jbowes:
And lowercase the y/n prompt, as the default in all cases is really
`no`.


promptui/prompt.go
```diff
@@ -24,8 +24,9 @@ type Prompt struct {
 	// characters.
 	Mask rune
 
-	IsConfirm bool
-	Preamble  *string
+	IsConfirm         bool
+	ConfirmDefaultYes bool
```
jeffandersen: Did you intend to use this?

jbowes: nope!

PR #112, opened 2016-12-01
Support usage of ** inside a pathexp being parsed

jeffandersen:
- Adds support for path inputs to contain `**`
- All paths will still be stored/used internally as their complete equivalent

### DIFF
--- a/pathexp/pathexp.go
+++ b/pathexp/pathexp.go
@@ -38,6 +38,7 @@ var (
 	slug           = regexp.MustCompile(`^` + slugstr + `$`)
 	globRe         = regexp.MustCompile(`^(` + slugstr + `)(\*?)$`)
 	fullglobOrGlob = regexp.MustCompile(`(^\*$)|(?:^(` + slugstr + `)(\*?)$)`)
+	doubleGlob     = regexp.MustCompile(`^\*\*$`)
 )
 
 const (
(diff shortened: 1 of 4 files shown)


jbowes:
you'll need to update the grammar def in the comment at the top of `pathexp.go`

pathexp/pathexp.go
```diff
@@ -38,6 +38,7 @@ var (
 	slug           = regexp.MustCompile(`^` + slugstr + `$`)
 	globRe         = regexp.MustCompile(`^(` + slugstr + `)(\*?)$`)
 	fullglobOrGlob = regexp.MustCompile(`(^\*$)|(?:^(` + slugstr + `)(\*?)$)`)
+	doubleGlob     = regexp.MustCompile(`^\*\*$`)
```
jbowes: wanna add some tests for doubleglob?

pathexp/pathexp.go
```diff
@@ -283,27 +284,51 @@ func New(org, project string, envs, services, identities, instances []string) (*
...
 		return nil, errors.New("path expressions must start with '/'")
 	}
+
 	// remove leading empty section
 	parts = parts[1:]
 
+	// identify double glob and replace remaining segments with glob
+	hasDoubleGlob := false
+	segments := make([]string, 6)
+	for i := range segments {
```
jbowes: double splatting needs to be greedy:`/jbowes/foo/**/default/foo` -> `/jbowes/foo/*/default/*/*/foo` (but should be `/jbowes/foo/*/*/*/default/foo`).

jbowes: two double splats should be invalid, as they'd be ambiguous:
`/jbowes/foo/**/default/**/foo` could be either:
- `/jbowes/foo/*/*/default/*/foo`
- `/jbowes/foo/*/default/*/*/foo`

jeffandersen:
Gonna roll this into some `torus ls` improvements, changing this opened a can of worms.

PR #116, opened 2016-12-02
Support Vim Movement Bindings via Readline

### LINKED ISSUE #56: support vim movement bindings
Hey guys, I think it would be cool if you could move around the choice selector with vim bindings (`hjkl`), this is what I initially tried to do and thought your product was broken.

bentranter:
Closes #56

Changes promptui to allow users to specify a boolean `vim` option in their `.torusrc`. Setting `vim = true` enables the Vim mode provided by [chzyer/readline](https://github.com/chzyer/readline).

It seems to work well so far, but there are two things I've found that aren't quite right:
  1. The cursor doesn't change when switching modes (normally not a problem, but when there's no way to tell what mode you're in, this could be a bit frustrating.
  2. You start in the equivalent of Vim's insert mode, which is confusing in any prompt where the user is required to select an option. Since you're in insert mode, you can't select an option until you leave insert mode by pressing [ESC] (or, by mashing your keyboard, like I did at first :P ).

I've set Vim mode to false by default, since it would definitely be confusing for new users.

---

###### Example

Run any command, ie, `torus-cli login` and try using Vim bindings to input things, and to navigate around your input.

###### Why the changes were made this way

I noticed readline had a Vim mode, so I added support in prompui for enabling the Vim mode.

In terms of implementation, I've gone with adding an `IsVimMode` field to the relevant structs in promptui, and populating them based on the in the user's preferences. This works, and is consistent, but doing

```go
preferences, err := prefs.NewPreferences()
if err != nil {
    return err
}
```

all the time feels repetitive. Maybe there's a better way, like reading from the preferences before running a command, and then passing that Vim config value as a parameter.

### DIFF
--- a/docs/commands/system.md
+++ b/docs/commands/system.md
@@ -15,11 +15,11 @@ The following are the available preferences:
 `core.public_key_file` | Location of the ed25519 public key file.
 `core.context` | Boolean determining if the `.torus.json` and defaults should be considered during command execution
 `core.auto_confirm` | Boolean determining if confirmation prompts should be automatically skipped (equivalent of always using `-y` command option)
+`core.vim` | Boolean determining if CLI input should use Vim bindings
 `defaults.org` | Organization name to be used with context
 `defaults.project` | Project name to be used with context
 `defaults.environment` | Environment name to be used with context
 `defaults.service` | Service name to be used with context
-`defaults.vim` | Boolean determining if CLI input should use Vim bindings
 
 ### set
 ###### Added [v0.1.0](https://github.com/manifoldco/torus-cli/blob/master/CHANGELOG.md)
--- a/docs/commands/system.md
+++ b/docs/commands/system.md
@@ -18,14 +18,15 @@ The following are the available preferences:
 `defaults.org` | Organization name to be used with context
 `defaults.project` | Project name to be used with context
 `defaults.environment` | Environment name to be used with context
-`defaults.service` | Service name to be used with context  
+`defaults.service` | Service name to be used with context
+`defaults.vim` | Boolean determining if CLI input should use Vim bindings
 
 ### set
 ###### Added [v0.1.0](https://github.com/manifoldco/torus-cli/blob/master/CHANGELOG.md)
 
 `torus prefs set <key> <value>` sets a preference value by key, where key is the dot-delimited path of the category and preference name.
 
-If value is omitted it defaults to boolean true.
+If value is omitted it defaults to boolean true, except for the Vim setting, which is false by default.
 
 ### list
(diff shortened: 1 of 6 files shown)


docs/commands/system.md
```diff
@@ -18,14 +18,15 @@ The following are the available preferences:
...
+`defaults.service` | Service name to be used with context
+`defaults.vim` | Boolean determining if CLI input should use Vim bindings
 
 ### set
 ###### Added [v0.1.0](https://github.com/manifoldco/torus-cli/blob/master/CHANGELOG.md)
 
 `torus prefs set <key> <value>` sets a preference value by key, where key is the dot-delimited path of the category and preference name.
 
-If value is omitted it defaults to boolean true.
+If value is omitted it defaults to boolean true, except for the Vim setting, which is false by default.
```
jbowes: 💯 Thank you for updating the docs!

prefs/prefs.go
```diff
@@ -58,6 +58,7 @@ type Defaults struct {
 	Project      string `ini:"project,omitempty"`
 	Environment  string `ini:"environment,omitempty"`
 	Service      string `ini:"service,omitempty"`
+	Vim          bool   `ini:"vim,omitempty"`
```
jbowes: The `vim` option would fit better into `core`. 
`defaults` Is reserved for code/organization structure.

promptui/select.go
```diff
@@ -14,8 +14,9 @@ const SelectedAdd = -1
 
 // Select represents a list for selecting a single item
 type Select struct {
-	Label string   // Label is the value displayed on the command line prompt.
-	Items []string // Items are the items to use in the list.
+	Label     string   // Label is the value displayed on the command line prompt.
+	Items     []string // Items are the items to use in the list.
+	IsVimMode bool     // Wether readline is using Vim mode.
```
jbowes: `Whether` :)

docs/commands/system.md
bentranter: No problem!

bentranter:
Thanks for the feedback @jbowes.

These should be fixed now:

- [x] Move the pref to core
- [x] Fix the whether typo
- [x] Default to, and always force, vim normal mode for select prompts

I didn't end up moving the prefs loading outside of `cmd/prompts.go`. I started moving some of them, but found that were a _lot_ of calls that needed to be changed, and a lot of spots where the preferences weren't already loaded, which meant adding another call to load the preferences. I figured it was a bit more readable (for me at least) to keep the logic that loads the preferences and the logic that uses those preferences in the same function. I can still make that change if you want, just figured I'd share my thoughts before making it.

jbowes:
Thanks @bentranter ! I've rebased, squashed (hope you don't mind), and merged your PR

bentranter:
No problem @jbowes!

PR #120, opened 2016-12-05
Support ** in paths, change torus ls behaviour to mimic ls more closely

jeffandersen:
See [command docs](https://github.com/manifoldco/torus-cli/blob/double-glob/docs/commands/secrets.md) for sample output.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,9 @@ _Unreleased_
   control levels of output in preparation for additional onboarding output.
 - Support vim movement bindings for interactive inputs. This can be enabled with
   `torus prefs set core.vim true`.
+- Support `**` in path expressions passed to commands.
+  `torus set /org/project/**/port 5000` is equivalent to `torus set /org/project/*/*/*/*/port`
+- `torus ls` behaviour changed to follow system `ls` more closely
 
 **Fixes**
 
--- a/api/projects.go
+++ b/api/projects.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"net/url"
 
+	"github.com/manifoldco/torus-cli/apitypes"
 	"github.com/manifoldco/torus-cli/identity"
 	"github.com/manifoldco/torus-cli/primitive"
 )
@@ -25,6 +26,15 @@ type projectCreateRequest struct {
 		OrgID *identity.ID `json:"org_id"`
 		Name  string       `json:"name"`
 	} `json:"body"`
+}
+
+// ProjectTreeSegment is the payload returned for a project tree
+type ProjectTreeSegment struct {
+	Org      *OrgResult              `json:"org"`
+	Envs     []*apitypes.Environment `json:"envs"`
+	Services []*apitypes.Service     `json:"services"`
+	Projects []*ProjectResult        `json:"projects"`
+	Profiles []*apitypes.Profile     `json:"profiles"`
 }
 
 // Create creates a new project with the given name within the given org
@@ -65,4 +75,18 @@ func (p *ProjectsClient) List(ctx context.Context, orgIDs *[]*identity.ID, names
 	projects := []ProjectResult{}
 	_, err = p.client.Do(ctx, req, &projects, nil, nil)
 	return projects, err
+}
+
+// GetTree returns a project tree
+func (p *ProjectsClient) GetTree(ctx context.Context, orgID *identity.ID) ([]ProjectTreeSegment, error) {
+	v := &url.Values{}
+	v.Set("org_id", orgID.String())
+	req, _, err := p.client.NewRequest("GET", "/projecttree", v, nil, true)
+	if err != nil {
+		return nil, err
+	}
+
+	segments := []ProjectTreeSegment{}
+	_, err = p.client.Do(ctx, req, &segments, nil, nil)
+	return segments, err
 }
--- a/cmd/ls.go
+++ b/cmd/ls.go
@@ -1,24 +1,34 @@
 package cmd
 
 import (
+	"context"
 	"fmt"
 	"sort"
+	"strings"
 
 	"github.com/urfave/cli"
 
 	"github.com/manifoldco/torus-cli/api"
+	"github.com/manifoldco/torus-cli/config"
 	"github.com/manifoldco/torus-cli/errs"
-	"github.com/manifoldco/torus-cli/identity"
 	"github.com/manifoldco/torus-cli/pathexp"
 )
+
+var targetMap = []string{"orgs", "projects", "envs", "services"}
 
 func init() {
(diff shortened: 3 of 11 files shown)


CHANGELOG.md
```diff
@@ -10,6 +10,8 @@ _Unreleased_
   control levels of output in preparation for additional onboarding output.
 - Support vim movement bindings for interactive inputs. This can be enabled with
   `torus prefs set core.vim true`.
+- Support `**` in path expressions passed to commands.
```
ianlivingstone: Can you provide an example in the changelog?

jeffandersen: Added

api/projects.go
```diff
@@ -27,6 +28,15 @@ type projectCreateRequest struct {
 	} `json:"body"`
 }
 
+// ProjectTreeSegment is the payload returns for an org tree
```
jbowes: this isn't right

api/projects.go
```diff
@@ -66,3 +76,17 @@ func (p *ProjectsClient) List(ctx context.Context, orgIDs *[]*identity.ID, names
 	_, err = p.client.Do(ctx, req, &projects, nil, nil)
 	return projects, err
 }
+
+// GetTree returns an org tree
```
jbowes: project tree

docs/commands/secrets.md
```diff
@@ -59,12 +59,49 @@ torus run -o example -- node ./bin/www --app api
...
-0 | All organizations| `torus ls /`
-1 | All projects within an org | `torus ls /org`
-2 | All environments (for the defined project) within an org | `torus ls /org/*`
-3 | All services (for the defined project) within an org | `torus ls /org/*/*`
-4 | All secrets within the defined path (org and path required) | `torus ls /org/project/*/*`
+Each level within the organization can be inspected by changing the segments supplied in the path. Wildcards cannot be used for the organization or project segments of the [path](../concepts/path.md).
+
+### Command Options
+
+  --recursive, -r | Show secrets for the expanded version of the supplied path
```
jbowes: this will need to be below the header now

cmd/ls.go
```diff
@@ -1,24 +1,39 @@
...
 	"github.com/urfave/cli"
 
 	"github.com/manifoldco/torus-cli/api"
+	"github.com/manifoldco/torus-cli/config"
 	"github.com/manifoldco/torus-cli/errs"
-	"github.com/manifoldco/torus-cli/identity"
 	"github.com/manifoldco/torus-cli/pathexp"
 )
 
+var targetMap = map[int]string{
```
jbowes: I don't think you gain anything using a map here. a slice with a bounds check during usage would be nicer IMO.

pathexp/pathexp.go
```diff
@@ -154,123 +194,59 @@ func segmentsEqual(a, b segment) bool {
 	}
 }
 
-// NewPartial creates a new path expression from the given path segments
+// New creates a new path expression from the given path segments
```
jbowes: I prefered the old way, with New/NewPartial, instead of the bool flag.

pathexp/pathexp.go
```diff
@@ -341,16 +377,31 @@ func (pe *PathExp) Org() string {
 	return string(pe.org)
 }
 
+// MatchProject identifies if the given string matches the project value
+func (pe *PathExp) MatchProject(str string) bool {
+	return matchSegment(pe.project, str)
+}
```
jbowes: What about doing the API like:

```go
func (pe *PathExp) Project() Segment { // or make project public
    return pe.project
}

type Segment interface {
    String() string
    Contains(other Segment) bool
}
```

Then you can do like:

```go
o, err := pathexp.ParseSegment("foo")
if pe.Project().Matches(o) {

}
```

ianlivingstone: I like this a lot!

jeffandersen: Yeah I didn't want to make all the fields public (since they weren't before) just to faciliatate compare.

cmd/ls.go
jeffandersen: Changed.

api/projects.go
```diff
@@ -27,6 +28,15 @@ type projectCreateRequest struct {
 	} `json:"body"`
 }
 
+// ProjectTreeSegment is the payload returns for a project tree
```
jbowes: *returned

pathexp/pathexp.go
```diff
@@ -51,26 +53,49 @@ const (
...
+	Envs       segment
+	Services   segment
+	Identities segment
+	Instances  segment
 }
 
 type segment interface {
 	String() string
+	Contains(subject string) bool
+	regexpString() string
```
jbowes: not needed as part of the interface anymore

pathexp/pathexp.go
```diff
@@ -51,26 +53,49 @@ const (
...
 type literal string
 type glob string
 type fullglob struct{}
 type alternation []segment
 
-func (l literal) String() string  { return string(l) }
-func (g glob) String() string     { return string(g) + "*" }
+func (l literal) String() string { return string(l) }
+func (l literal) Contains(subject string) bool {
+	return matchSegmentRegex(l.regexpString(), subject)
```
jbowes: this can just be string(l) == subject

pathexp/pathexp.go
```diff
@@ -51,26 +53,49 @@ const (
...
+}
+func (l literal) regexpString() string {
+	return `^` + l.String() + `$`
+}
+
+func (g glob) String() string { return string(g) + "*" }
+func (g glob) Contains(subject string) bool {
+	return matchSegmentRegex(g.regexpString(), subject)
```
jbowes: This can be `strings.Index(subject, (string(g)) == 0`

pathexp/pathexp.go
```diff
@@ -51,26 +53,49 @@ const (
...
+}
+func (g glob) regexpString() string {
+	return `^` + g.String() + `(` + slugstr + `)?$`
+}
+
 func (f fullglob) String() string { return "*" }
+func (f fullglob) Contains(subject string) bool {
+	return matchSegmentRegex(f.regexpString(), subject)
```
jbowes: this is always `true`

pathexp/pathexp.go
```diff
@@ -84,6 +109,21 @@ func (a alternation) String() string {
 
 	return "[" + strings.Join(strs, "|") + "]"
 }
+func (a alternation) Contains(subject string) bool {
+	return matchSegmentRegex(a.regexpString(), subject)
```
jbowes: This can be
```go
for _, s := range a {
    if s.Contains(subject) {
        return True
    }
}
````

PR #134, opened 2016-12-06
Bugfixes for torus ls

jeffandersen:
1. We were not surfacing the proper error object, so projecttree lookup would panic when the org supplied was unknown to the user
2. Project and org must be valid segments for certain (`ls` specific) lookups, because they are `type literal` in pathexp, perform validation inside `torus ls`
3. Wildcard match the literal segments properly (previously it was attempting `org*==org` exact comparison). Extended this matching to secret names inside `torus ls` as they are not part of the `pathexp` object.
4. Remove recursive flag. `-r` was originally added to show the secret values for a partial path, but with `**` and a design decision to make `torus ls` require a path/be deliberate, it doesn't make sense anymore.

No changelog added as these are unreleased.


cmd/ls.go
```diff
@@ -23,24 +23,11 @@ func init() {
 		Usage:     "Explore all objects your account has access to",
 		Category:  "SECRETS",
 		Flags: []cli.Flag{
-			recursiveFlag(),
```
jbowes: This will need to be removed from the docs page too

cmd/ls.go
```diff
@@ -23,24 +23,11 @@ func init() {
...
-			newSlicePlaceholder("environment, e", "ENV", "Use this environment.",
-				"*", "TORUS_ENVIRONMENT", false),
-			newSlicePlaceholder("service, s", "SERVICE", "Use this service.",
-				"default", "TORUS_SERVICE", false),
-			newSlicePlaceholder("user, u", "USER", "Use this user (identity).",
-				"*", "TORUS_USER", false),
-			newSlicePlaceholder("machine, m", "MACHINE", "Use this machine.",
-				"*", "TORUS_MACHINE", false),
-			newSlicePlaceholder("instance, i", "INSTANCE", "Use this instance.",
-				"*", "TORUS_INSTANCE", false),
```
jbowes: Removing this args should be in the changelog imo

cmd/ls.go
```diff
@@ -172,29 +173,35 @@ func listObjects(ctx *cli.Context) error {
...
-	defined := 0
-	path := "/"
-	if len(args) < 1 {
-		if recursive {
-			return path, "secrets"
-		}
-		return path, "orgs"
-	}
-
+func identifyTarget(args []string, recursive bool) (*pathexp.PathExp, string, error) {
```
jbowes: A zero length string will cause this func to panic, ie

`torus ls ""`

cmd/ls.go
jeffandersen: Added.

jeffandersen:
Fixed secrets listing, was due to `**`

```
$ torus ls "/jeff0190rc/awesome/**"
/jeff0190rc/awesome/dev-jeff0190rc/default/jeff0190rc/1/port
$ torus ls "/jeff0190rc/awesome/**/*"
/jeff0190rc/awesome/dev-jeff0190rc/default/jeff0190rc/1/port
$ torus ls "/jeff0190rc/awesome/**/nope"
```

jeffandersen:
Secrets listing works now.

```
$ torus ls "/jeff0190rc/awesome/**"
/jeff0190rc/awesome/dev-jeff0190rc/default/jeff0190rc/1/port
$ torus ls "/jeff0190rc/awesome/**/*"
/jeff0190rc/awesome/dev-jeff0190rc/default/jeff0190rc/1/port
$ torus ls "/jeff0190rc/awesome/**/nope"
```

PR #146, opened 2016-12-10
ObserverTest: prevent a deadlock situation.

jelmersnoeck:
The tests described in the test file didn't really describe a real world
situation. In the real world, an observer is either always present and
started well before we send notifications or no observer will be ever
present at all.

Running the tests, a deadlock occurred sometimes, locking up the tests
and eventually failing them. This was because the Observer in our tests
didn't register itself in time before we created a new notification. The
notification then got sent through the system and dropped because no
observer was present. Once the observer then started, it would keep listening
until the notification came through, resulting in a deadlock since the
event was already dropped beforehand.

This patch does actually not fix the deadlock situation itself, it is
still possible to happen - there is however logging present now to
document this. We do however set up the tests more reliable in a way
that we expect the real world to behave. We ensure that the observer is
registered before we send the notification. This way once we send a
notification, the observer sends it pack to the ResponseRecorder.

This PR also takes care of a data race:

`*notifier.Notify()` can be used concurrently. Because of this behaviour, it can
trigger a race condition when two events happen at the same time where
one would incremment the current counter and one would read from it.

This implements a Read/Write lock which allows this behaviour by locking
the instance on updating the counter and only adding a Read lock when
reading the value. This allows us to only write from a single point at a
single moment in time, but read from multiple points at a single moment
in time. It will also prevent users from reading the counter when
someone else is writing to it.

### DIFF
--- a/daemon/observer/observer.go
+++ b/daemon/observer/observer.go
@@ -125,7 +125,8 @@ func (n *Notifier) Notifier(total uint) *Notifier {
 	return notifier
 }
 
-// Notify publishes an event to all SSE observers.
+// Notify publishes an event to all SSE observers. This function panics when it
+// is called more often than it is supposed to have been called.
 func (n *Notifier) Notify(eventType EventType, message string, increment bool) {
 	notif := &notification{
 		Type:      eventType,
--- a/daemon/observer/observer.go
+++ b/daemon/observer/observer.go
@@ -10,6 +10,7 @@ import (
 	"fmt"
 	"log"
 	"net/http"
+	"sync"
 )
 
 type ctxkey string
@@ -76,6 +77,7 @@ type Notifier struct {
 	notifications  chan<- *notification
 	observerClosed <-chan int
 	ctxDone        <-chan struct{}
+	sync.RWMutex
 }
 
 func (t *transaction) start() {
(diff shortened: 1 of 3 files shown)


daemon/observer/observer.go
```diff
@@ -132,13 +135,17 @@ func (n *Notifier) Notify(eventType EventType, message string, increment bool) {
...
 
 	if increment {
+		n.Lock()
 		n.current++
+		n.Unlock()
 	}
 
+	n.RLock()
 	if n.current > n.total {
 		panic(fmt.Sprintf(
```
jbowes: What happens to the lock if this panic occurs?

daemon/observer/observer_test.go
```diff
@@ -153,3 +157,87 @@ func TestObserverServeHTTP(t *testing.T) {
...
+
+		go o.ServeHTTP(rw, r)
+		<-rw.flushed
+
+		go n.Notify(Progress, "hi", true)
+		<-rw.flushed
+	}
+}
+
+// This benchmarks ensures the concurrent behaviour of the notifier. We should
```
jbowes: `benchmark` :)

daemon/observer/observer.go
jelmersnoeck: I have changed this as this could indeed cause locking issues seeing that the panic is being recovered higher up the chain.

It now reads the data into two variables, isolated in a single rlock/runlock.

PR #147, opened 2016-12-11
Put worklog item type logic into their own types

jbowes:
Extract out the `List` and `Resolve` logic for each worklog item type
into new structs that implement a common `handler` interface.

This keeps the logic for each type distinct, which should (hopefully)
make for easier reading, and easier addition of new item types.

With the logic now distinct and behind a map lookup, take advantage of
this during `Get` and `Resolve` to extract the item type from the
requested ID, and only lookup that type of item, bypassing all the
others.

### DIFF
--- a/apitypes/worklog.go
+++ b/apitypes/worklog.go
@@ -17,6 +17,8 @@ const (
 	SecretRotateWorklogType WorklogType = 1 << iota
 	MissingKeypairsWorklogType
 	InviteApproveWorklogType
+
+	AnyWorklogType WorklogType = 0xff
 )
 
 // WorklogResultType is the string type of worklog results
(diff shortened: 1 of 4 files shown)


apitypes/worklog.go
```diff
@@ -17,17 +17,19 @@ const (
...
 type WorklogResultType string
 
 // WorklogResult result states.
 const (
-	SuccessWorklogResult = "success"
-	FailureWorklogResult = "failure"
-	ErrorWorklogResult   = "error"
-	ManualWorklogResult  = "manual"
+	SuccessWorklogResult WorklogResultType = "success"
+	FailureWorklogResult WorklogResultType = "failure"
```
ianlivingstone: 👍

daemon/logic/worklog.go
```diff
@@ -19,19 +20,95 @@ import (
...
 type Worklog struct {
-	engine *Engine
+	engine   *Engine
+	handlers map[apitypes.WorklogType]worklogTypeHandler
+}
+
+func newWorklog(e *Engine) Worklog {
+	w := Worklog{
+		engine: e,
+		handlers: map[apitypes.WorklogType]worklogTypeHandler{
```
ianlivingstone: I like this a lot 👍

PR #152, opened 2016-12-13
Teach the worklog how to correct keyring memberships

jbowes:
If an invite or machine creation fails partway through, we can end up in
a state where the user/machine is created, but are not given access to
all keyrings, and the secrets therein.

Teach the worklog how to identify and resolve this problem.

### DIFF
--- a/daemon/logic/worklog.go
+++ b/daemon/logic/worklog.go
@@ -41,6 +41,7 @@ type worklogTypeHandler interface {
 	list(context.Context, *envelope.Unsigned) ([]apitypes.WorklogItem, error)
 	resolve(context.Context, *observer.Notifier, *identity.ID,
 		*apitypes.WorklogItem) (*apitypes.WorklogResult, error)
+	resolveErr() string
 }
 
 // List returns the list of all outstanding worklog items for the given org
@@ -100,11 +101,32 @@ func (w *Worklog) Resolve(ctx context.Context, n *observer.Notifier,
 		return nil, nil
 	}
 
-	return w.handlers[item.Type()].resolve(ctx, n, orgID, item)
+	handler := w.handlers[item.Type()]
+	result, err := handler.resolve(ctx, n, orgID, item)
+
+	// We handle errors from the handler's resolve differently than regular
+	// errors; for every other part of the daemon code they're non-errors, sent
+	// back to the caller for display. The caller can then continue trying to
+	// resolve other worklog items.
+	if err != nil {
+		result = &apitypes.WorklogResult{
+			ID:      item.ID,
+			State:   apitypes.ErrorWorklogResult,
+			Message: handler.resolveErr() + ": " + err.Error(),
+		}
+	}
+
+	return result, nil
 }
 
 type secretRotateHandler struct {
 	engine *Engine
+}
+
+func (secretRotateHandler) resolveErr() string {
+	// This won't happen, because rotation must be manual. Let's include an
+	// error message just in case, though!
+	return "Error rotating secret"
 }
 
 func (h *secretRotateHandler) list(ctx context.Context, org *envelope.Unsigned) ([]apitypes.WorklogItem, error) {
@@ -166,6 +188,10 @@ type missingKeypairsHandler struct {
 	engine *Engine
 }
 
+func (missingKeypairsHandler) resolveErr() string {
+	return "Error generating keypairs"
+}
+
 func (h *missingKeypairsHandler) list(ctx context.Context, org *envelope.Unsigned) ([]apitypes.WorklogItem, error) {
 	encClaimed, sigClaimed, err := fetchRegistryKeyPairs(ctx, h.engine.client, org.ID)
 	if err != nil {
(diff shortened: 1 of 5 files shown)


ianlivingstone:
Also needed for re-encoding users into keyrings in teh case of account recovery!

daemon/logic/worklog.go
```diff
@@ -100,13 +107,30 @@ func (w *Worklog) Resolve(ctx context.Context, n *observer.Notifier,
 		return nil, nil
 	}
 
-	return w.handlers[item.Type()].resolve(ctx, n, orgID, item)
+	handler := w.handlers[item.Type()]
+	result, err := handler.resolve(ctx, n, orgID, item)
+
+	if err != nil {
+		result = &apitypes.WorklogResult{
```
ianlivingstone: Why return a special result for an error instead of a standard error?

ianlivingstone: Ah I see, it was here before, you jsut cleaned it up a bit. Still a question though! Perhaps add a comment?

daemon/logic/worklog.go
```diff
@@ -242,17 +270,257 @@ func (h *inviteApproveHandler) resolve(ctx context.Context, n *observer.Notifier
...
 	if err != nil {
-		return &apitypes.WorklogResult{
-			ID:      item.ID,
-			State:   apitypes.ErrorWorklogResult,
-			Message: "Error approving invite: " + err.Error(),
-		}, nil
+		return nil, err
+	}
+
+	// Find all of the credential graphs in this org.
```
ianlivingstone: You should be able to just `GET /keyrings` ?

ianlivingstone: https://github.com/manifoldco/registry/blob/master/schema/request/keyrings/list/response.json

Returns everything *but* the credentials -- which we don't need!

ianlivingstone: Thoughts? Perhaps you wanted to standardize on the the credentialgraph endpoint for caching?

daemon/logic/worklog.go
```diff
@@ -242,17 +270,257 @@ func (h *inviteApproveHandler) resolve(ctx context.Context, n *observer.Notifier
...
+	//
+	// The paths map takes care of eliminating multiple versions of the keyring;
+	// the subsequent List call will return all versions.
+	cgs := newCredentialGraphSet()
+	orgName := org.Body.(*primitive.Org).Name
+	paths := make(map[string]*pathexp.PathExp)
+	for _, project := range projects {
+		projName := project.Body.(*primitive.Project).Name
+		graphs, err := h.engine.client.CredentialGraph.Search(ctx,
+			"/"+orgName+"/"+projName+"/*/*/*/*", h.engine.session.AuthID())
```
ianlivingstone: ugh - i really wan to fix this tech debt ;(

daemon/logic/worklog.go
```diff
@@ -242,17 +270,257 @@ func (h *inviteApproveHandler) resolve(ctx context.Context, n *observer.Notifier
...
+	// members of a keyring based on ACLs.
+	members, err := getKeyringMembers(ctx, h.engine.client, org.ID)
+	if err != nil {
+		return nil, err
+	}
+
+	missing := make(map[string]apitypes.WorklogItem)
+	for _, graph := range graphs {
+		for _, member := range members {
+			m, _, err := graph.FindMember(&member)
```
ianlivingstone: does this handle revoked keys?

ianlivingstone: i.e. it says a user isn't a member if the included keys are revoked?

ianlivingstone: (If you have a valid keypair but the currently encoded key is revoked, we should encode your new keys)

daemon/logic/worklog.go
jbowes: ya can do. it's just to differentiate between a fatal error in the daemon/registry vs an error in an individual worklog item; so if you're resolving many at once, you can keep going.

daemon/logic/worklog.go
jbowes: which part? ranging over projects?
The api endpoint says it should support `*` project name, but doesn't :(

daemon/logic/worklog.go
jbowes: this will take a lil' bit. i'll push up the other changes now.

daemon/logic/worklog.go
ianlivingstone: Right - forgot about the "activity" component.

daemon/logic/worklog.go
ianlivingstone: That exactly! It wouldn't be a ton of work to push this behaviour into the server, and we do this in many different places (torus ls).

daemon/logic/worklog.go
ianlivingstone: Ok - we can do it as a separate PR as well -- we don't have a good mechanism for revoking keypairs for testing either, perhaps we should implement a hidden `torus keypairs revoke` until we have a full fledged account recovery flow?

jbowes: yeah i'll do it in another PR. lets get this party moving. https://github.com/manifoldco/torus-cli/issues/153

Issue #153, opened 2016-12-13
teach worklog keyring adjustment about revoked keypairs


jbowes:
When correcting keyring memberships, the worklog should know about when a user's keypair was revoked and replaced with a new one. This should trigger:
- adding the new keypair into the existing active keyrings (so the user has access)
- in the secrets worklog handler - rotating the existing keyring versions (happens already)

For testing, it's probably time we add a hidden `keypairs revoke` command.

jbowes:
closed by #162

PR #158, opened 2016-12-14
Introduce the hidden keypairs revoke command

jbowes:
This command is purely for testing and debugging; real revocation will
come packaged in a nicer form, bundled with other actions.

Along with this, teach `keypairs list` to display the actual validity
state of a keypair.

related: #153

### DIFF
--- a/apitypes/claimtree.go
+++ b/apitypes/claimtree.go
@@ -1,9 +1,15 @@
 package apitypes
 
 import (
+	"errors"
+
 	"github.com/manifoldco/torus-cli/envelope"
 	"github.com/manifoldco/torus-cli/primitive"
 )
+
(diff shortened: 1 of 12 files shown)


apitypes/claimtree.go
```diff
@@ -1,10 +1,16 @@
...
 import (
+	"errors"
+
 	"github.com/manifoldco/torus-cli/envelope"
 	"github.com/manifoldco/torus-cli/primitive"
 )
 
+// ErrClaimCycleFound is returned when a cycle is found within the claims.
+// this *should* be impossible, as they are signed.
+var ErrClaimCycleFound = errors.New("Cycle detected in signed claims")
```
ianlivingstone: !!! nice

PR #162, opened 2016-12-15
Worklog keyring regenerated keys

jbowes:
Only return unrevoked keyring memberships

When looking for a keyring membership, skip over any memberships that
match the user/machine token, but have been revoked; only return active
unrevoked memberships.

With this, the worklog can detect when a user was once in a keyring, had
their membership revoked, and then need to be re-added.


ianlivingstone:
Is there a way for us to test this through worklog as a part of our QA flow?

jbowes:
> Is there a way for us to test this through worklog as a part of our QA flow?

absolutely! i'll add some sections

PR #169, opened 2016-12-16
Create github releases when releasing

jbowes:
Tie our tags to proper github releases. Include the relevant changelog
entry if it exists. Mark -rc releases as prerelease.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,13 @@
 # CHANGELOG
 
+## v0.22.0
+
+_Unreleased_
+
+**Notable Changes**
+
+- Publish release details to GitHub as proper releases.
+
 ## v0.21.0
 
 _2016-12-16_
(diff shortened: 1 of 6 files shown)


jbowes:
I'm waffling on if we should create proper prerelease flagged releases for `rc` tags, or just leave them as tags. Which way will be the least obtrusive in the releases list? We could try and put changes between `rcs` in the  release body, but since we `rc` so fast, I don't see much of a point to that.

No `releases`:
image

prerelease `releases`:

image

(imagine a bunch more prerelease entries before the next release)

ianlivingstone:
No releases would be the least jntrusive I think

jbowes:
> No releases would be the least jntrusive I think

👍 I'll flush the existing releases, only fill in the non rc ones, and update this pr.

ianlivingstone:
Can you update the release instructions as well, since we'll need GH creds to release!

jbowes:
> Can you update the release instructions as well, since we'll need GH creds to release!

done

PR #170, opened 2016-12-18
typed envelopes

jbowes:
Generate `envelope.$Type` for every primitive type. Use these instead of `envelope.[Signed|Unsigned]` (eliminating type assertions). Remove the now redundant types in `api` and `apitypes`.


ianlivingstone:
Amaze

jbowes:
Plan for this:
- Get a nice interface in place for the session/auth stuff (`XXX` included in changes)
- Reduce type duplication that will now exist between api/apitypes and envelope
- Share api clients between frontend and daemon
- (Maybe) Replace envelope/primitive split with a flattened primitive only struct that will serialize and deserialize properly
- Interfaces and embedded shared structs all over the place
- test it? maybe. :P

jbowes:
Ok i'm gonna stop here for now; this pr is big enough :) now to make sure it works!

daemon/registry/credentialgraph.go
```diff
@@ -129,10 +118,10 @@ func (c *CredentialGraphClient) getGraph(ctx context.Context, query url.Values)
 	}
 
 	resp := []struct {
-		Keyring     *envelope.Signed  `json:"keyring"`
-		Members     json.RawMessage   `json:"members"`
-		Credentials []envelope.Signed `json:"credentials"`
-		Claims      []envelope.Signed `json:"claims"`
+		Keyring     *envelope.Signed              `json:"keyring"`
+		Members     json.RawMessage               `json:"members"`
+		Credentials []envelope.Signed             `json:"credentials"`
```
ianlivingstone: I guess you an't unpack it until *later* on

ianlivingstone:
Please open issues for the other sections.

daemon/registry/credentialgraph.go
jbowes: yeah i'm just reusing `Signed` here as a generic container before finding the correct struct.

jbowes:
> Please open issues for the other sections.

done! I opend #171 to track it al.

PR #174, opened 2016-12-22
Unified api groundwork

jbowes:
This pr contains the groundwork needed to begin deduplicating and centralizing the two api trees:
- disable `log` output for regular cli commands (so we can keep log statements in the registry api code)
- Use a shared interface and implementation for http roundtripping
- move `daemon/registry` to `registry`
- embed registry.Client in api.Client. From here, the next PR can start removing implementations from api.Client and allow the underlying registry.Client implementation to kick in.


ianlivingstone:
LGTM

PR #177, opened 2016-12-27
Remove Signup Object

ianlivingstone:
The server now supports an optional state field being supplied during
signup, so we can remove the duplication inside the CLI code base!


registry/users.go
```diff
@@ -121,18 +120,3 @@ func validateSelf(s *envelope.User) error {
 
 	return nil
 }
-
-type omit struct{}
```
jbowes: This should have been `type omit *struct{}` :(

I like what you've done muuuch better though

PR #183, opened 2016-12-29
Introduce v2 User w/ Upgrade from v1->v2

ianlivingstone:
The version 2 user schema uses passphrase derived public key
authentication for creating auth tokens.

New signups will be created as v2 Users, while v1 Users will be upgraded
to v2 during the authentication process.

Once a user has upgraded to v2, the old (now deprecated) hmac based
authentication flow will no longer be allowed.

Cliff notes of changes made:

* Introduction of UserInf interface (common access layer for v1 and v2 Users)
* Convenience method for accessing the underlying `byte` array for the
  base64 package.
* Bump to supporting Registry API version `0.3.0`
* Introduction of the `Token` primitive type and envelope

### DIFF
--- a/daemon/logic/machine.go
+++ b/daemon/logic/machine.go
@@ -39,7 +39,7 @@ func (m *Machine) CreateToken(ctx context.Context, notifier *observer.Notifier,
 		return nil, err
 	}
 
-	keypair, err := crypto.DeriveLoginKeypair(ctx, secret, salt)
+	keypair, err := crypto.DeriveLoginKeypair(ctx, *secret, salt)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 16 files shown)


daemon/logic/machine.go
```diff
@@ -39,7 +39,7 @@ func (m *Machine) CreateToken(ctx context.Context, notifier *observer.Notifier,
 		return nil, err
 	}
 
-	keypair, err := crypto.DeriveLoginKeypair(ctx, secret, salt)
+	keypair, err := crypto.DeriveLoginKeypair(ctx, secret.ByteArray(), salt)
```
jbowes: `*secret` :)

primitive/primitive.go
```diff
@@ -51,6 +57,19 @@ type User struct { // type: 0x01
...
+type User struct { // type: 0x01
+	BaseUser
+	v2Schema
+	PublicKey *LoginPublicKey `json:"public_key"`
+}
+
+// UserV1 is the body of a user object
+type UserV1 struct { // type: 0x01
+	BaseUser
+	v1Schema
```
jbowes: To fit the style of the other structs put the v*Schema first, please.

And for now, include `mutable` on the User and UserV1 types, not the base one (again just to match the others)

PR #185, opened 2017-01-06
Don't create keyring worklog entries for keypairless users

jbowes:
If a user is missing access to a keyring, but they don't have a valid,
unrevoked keypair, don't generate worklog items for other users to add
them to the keyring; they won't be able to until the user has a valid
keypair.

### DIFF
--- a/daemon/logic/worklog.go
+++ b/daemon/logic/worklog.go
@@ -335,6 +335,11 @@ func (h *keyringMembersHandler) list(ctx context.Context, org *envelope.Org) ([]
 		return nil, err
 	}
 
+	claimTrees, err := h.engine.client.ClaimTree.List(ctx, org.ID, nil)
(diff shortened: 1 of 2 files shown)


daemon/logic/worklog.go
```diff
@@ -354,6 +359,15 @@ func (h *keyringMembersHandler) list(ctx context.Context, org *envelope.Org) ([]
 				continue
 			}
 
+			// We now know the user/machine should have access to this keyring,
+			// but at the moment they do not. See if they do have a valid
+			// encryption key we can use to add them to the keyring.
+			targetPubKey, _ := findEncryptionPublicKey(claimTrees, org.ID, &member)
+			if targetPubKey == nil { // nothing we can do for this user right now
+				continue
+
```
ianlivingstone: empty space?

PR #187, opened 2017-01-15
Updates

federicoruggi:
Added the capability to fetch updates from `https://get.torus.sh/manifest.json` and inform user via the cli about new versions.

Such feature can be toggled with the `check_updates` prefs variable.

It's meant to be added as chained command via the cli middleware, it's added for demo purposes in the `login` and `version` commands.


jbowes:
Looking good so far!

Mind signing the CLA?

federicoruggi:
All your comments make sense, see the updated PR with the new commits ;)

PR #205, opened 2017-02-03
Round tripper

jeffandersen:
Change the http client to use RoundTripper in preparation for future client-side use.


jbowes:
https://www.youtube.com/watch?v=othSH7Bh6d0

Issue #123, opened 2016-12-06
Windows Support


ianlivingstone:
We've had a lot of questions regarding windows, it'd be great if we could add proper support. The largest change will probably be handling the Daemon socket which currently uses a Unix file socket.

There is other work as well, to acclimatize the environment to a windows user. We'll also need a decent strategy for ensuring the CLI tool continues to work on Windows through CI.

ianlivingstone:
@jbowes You've done some thinking on this topic, can you add color?

jbowes:
![image](https://cloud.githubusercontent.com/assets/12997/20926236/a8bad870-bb91-11e6-9bc3-115e2178b284.png)

jbowes:
It will certainly be worth supporting the windows subsystem for linux, but hopefully that will just work, and all we'll need to do is document it.

What we're interested in here is properly supporting usage of torus compiled as a windows binary, running under `cmd.exe` (and maybe powershell, too?).

TODO:
- [ ] proper session backgrounding of the daemon for windows (we use `setsid(2)` on *nix which has direct equivalent on windows
- [ ] domain socket equivalent with proper permission setup
- [ ] sanitization of directory path lookup/creation; home dir lookups
- [ ] make sure 32 bit works? my gut feeling is there's still a lot of 32 bit usage out there for windows
- [ ] windows CI setup

jelmersnoeck:
My plan of action for this looks as follows (for now):

- [ ] Get the daemon working (start/status/stop)
- [ ] Get basic account functions working (login, logout)

This will bring up general issues that will be widespread throughout the system for Windows (paths, file permissions etc.). Once login/logout works, it means we can also communicate with the server.

Having briefly looked into this, this will most likely mean we'll have a lot of windows specific build files.

The next step will be to get all tests working on a Windows VM. This will just be running the tests, see what and why they fail and tackle them one by one. A proper Windows CI setup will help with this as well.

- [ ] Setup Windows CI
- [ ] Get all tests passing on Windows

ianlivingstone:
> Having briefly looked into this, this will most likely mean we'll have a lot of windows specific build files.

How are we going to organize these? I'm not familiar myself with how to organize platform specific implementations in Go.

jelmersnoeck:
@ianlivingstone as an example:

https://github.com/golang/go/blob/master/src/net/file_unix.go
https://github.com/golang/go/blob/master/src/net/file_windows.go

[...]

PR #207, opened 2017-02-07
Add Windows support.

jelmersnoeck:
This is the first step towards having Torus working on Windows, as per [#123](https://github.com/manifoldco/torus-cli/issues/123).

This patch splits up some unix specific behaviour and adds build tags so we can
use a Windows specific implementation:

- use Named Pipes on Windows (only) as a net connection.
- disable the prompt styler on Windows
- we don't set `Setsid: true` on the daemon command for Windows
- we set torusRoot according to `HomePath` and prepend `HOMDERIVE` on Windows
- per OS `rootPathLength` to decide if we're on a root path or not
- set the transport address per OS, encapsulate it's behaviour and rename `SocketPath` to `TransportAddress`

Needs attention:

- disabled prompt for Windows


jelmersnoeck:
@ianlivingstone @jbowes I've implemented npipes, which was surprisingly easy. I'm going to read up on the documentation a little bit more for named pipes in general to dig through all the security implementations, but from an initial look, this seems like the perfect solution (also, Keybase are using this).

One suggestion about this, it might be worth it implementing the `DialTimeout` setup, this way we ensure we don't hang forever and close connections when they take too long. I haven't done this here to mimic the same behaviour as the regular setup, but we could do this in a separate PR.

I'll also have a look if we can get around the `setsid` for Windows in a way.

ianlivingstone:
> One suggestion about this, it might be worth it implementing the DialTimeout setup, this way we ensure we don't hang forever and close connections when they take too long. I haven't done this here to mimic the same behaviour as the regular setup, but we could do this in a separate PR.

Makes tons of sense, want to split off another issue to set this up?

jelmersnoeck:
So regarding `Setsid`, apparently this is the default on Windows and we don't need to set it.

(taken from the [Keybase documentation](https://github.com/keybase/client/blob/1d41ead7d6b2abc9e5242e94bf54bf7966d91240/go/client/fork_server_windows.go#L44-L46))

ianlivingstone:
> So regarding Setsid, apparently this is the default on Windows and we don't need to set it.

Nice!

ianlivingstone:
What's the installation process look like for someone on Windows? Should we split off another issue for setting that up? 

Do they just put the `torus.exe` somewheres in their path? I suppose we need some documentation surrounding that as well.

ianlivingstone:
We're just waiting on:

- promptui changes
- creating a `.zip` of `torus.exe`
- Installation instructions
- Creation of an issue to make an `.msi` installer

Before this can be merged.

jelmersnoeck:
With regards to the UTF8 characters, I've used simple latin fallbacks on Windows. By default, Windows doesn't support UTF8 in the prompt.

When we use a `.msi` installer, we could potentially look at turning this on by default?

http://superuser.com/questions/269818/change-default-code-page-of-windows-console-to-utf-8/269857#269857

api/transport_windows.go
```diff
@@ -0,0 +1,17 @@
+package api
```
ianlivingstone: We might have discussed this previously, but my memory is foggy, is their a reason we only use npipe on windows and not for both?

appveyor.yml
```diff
@@ -0,0 +1,31 @@
+version: "{build}"
```
ianlivingstone: :1st_place_medal:

appveyor.yml
```diff
@@ -0,0 +1,31 @@
...
+  - go env
+  - appveyor DownloadFile https://github.com/Masterminds/glide/releases/download/v%GLIDEVERSION%/glide-v%GLIDEVERSION%-windows-amd64.zip
+  - md %GOPATH%\bin
+  - 7z e glide-v%GLIDEVERSION%-windows-amd64.zip -o%GOPATH%\bin *.exe -r
+
+build_script:
+  - c:\gopath\bin\glide.exe install
+
+test_script:
+  - FOR /f "" %%G in ('go list ./... ^| find /i /v "/vendor/"') DO (go test -run=. -bench=. -v %%G)
```
ianlivingstone: Is there a reason we can't just run `make test` here? Or is it due to the fact that `make` isn't available.

config/socket_windows.go
```diff
@@ -0,0 +1,5 @@
+package config
+
+func setTransportAddress(cfg *Config) {
+	cfg.TransportAddress = `\\.\pipe\manifoldco\torus`
```
ianlivingstone: Is this relative to the user, or to the system as a whole?

daemon/socket/socket_windows.go
```diff
@@ -0,0 +1,13 @@
...
+package socket
+
+import (
+	"net"
+
+	"github.com/natefinch/npipe"
+)
+
+func makeSocket(transportAddress string, groupShared bool) (net.Listener, error) {
+	return npipe.Listen(transportAddress)
```
ianlivingstone: Perhaps we should add a comment re: windows pipe permissions and how this is only accessible to the user for future reference?

api/transport_windows.go
jelmersnoeck: Good point! We could just use it for all platforms. I'll change this back to just use npipe.

daemon/socket/socket_windows.go
jelmersnoeck: 👍

appveyor.yml
jelmersnoeck: By default `make` isn't available. There's a [Windows variant](http://gnuwin32.sourceforge.net/packages/make.htm) available but we'd have to install it. (I haven't tested this, so not sure how compatible it is with our Makefile)

config/socket_windows.go
jelmersnoeck: From what I get from [the docs](https://msdn.microsoft.com/en-us/library/windows/desktop/aa365783(v=vs.85).aspx) this is system wide. Thinking back now, this probably goes against what we're trying to achieve with Unix Sockets.

Looking back at how Keybase do it, they make a string (`manifoldco\torus` in our case) based on the homedir etc. This makes it unique but still "guessable". I'll retest this again because re-reading the docs, it shouldn't work with backslashes.

api/transport_windows.go
jelmersnoeck: Actually, looking back at this. The npipe package is only for Windows (only builds for Windows), which makes sense as it's a Windows feature. So we'd have to differentiate after all.

config/socket_windows.go
jelmersnoeck: Fixed the backslash. Also confirmed it's [being created](http://screenshots.siphoc.com/2017-03-11-14-19-47.png) now.

api/transport_windows.go
ianlivingstone: 👍

config/socket_windows.go
ianlivingstone: Let's just add some documentation to the README as a caveat for windows support re: system wide. Then we can get this merged up.

ianlivingstone: Also, let's split off an issue to investigate/pursue further when we have time.

PR #214, opened 2017-03-05
Transport timeouts.

### LINKED ISSUE #210: Use DialTimeout for the transport settings.
By implementing `DialTimeout` instead of `Dial` for our Socket Dial, we ensure that we don't hang forever.

Currently, if there's an issue, it could be that the dial just "hangs". By enforcing a timeout, we make sure it stops after a certain amount of time.

Seeing that it's a local Dial, I'd suggest a timeout of 1s.

jelmersnoeck:
This implements a patch to address issue #210.

I've set up a 1 second timeout for the Dial and a 30s timeout for actual requests.

@jbowes @ianlivingstone do you think 30s is good enough for actual requests or should we consider something else?

**Note:** this depends on #207 to be merged in. [This](https://github.com/jelmersnoeck/torus-cli/compare/jelmer/123-windows-support...jelmersnoeck:jelmer/210-transport-timeout?expand=1) is the actual diff whilst that one isn't merged in yet.

closes #210

### DIFF
--- a/api/client.go
+++ b/api/client.go
@@ -6,6 +6,7 @@ import (
 	"encoding/json"
 	"net/http"
 	"net/url"
+	"time"
 
 	"github.com/donovanhide/eventsource"
 	"github.com/satori/go.uuid"
@@ -14,7 +15,10 @@ import (
 	"github.com/manifoldco/torus-cli/registry"
 )
 
-const daemonAPIVersion = "v1"
+const (
+	daemonAPIVersion = "v1"
+	dialTimeout      = time.Second
+)
 
 type blacklisted struct{}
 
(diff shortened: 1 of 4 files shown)


ianlivingstone:
> @jbowes @ianlivingstone do you think 30s is good enough for actual requests or should we consider something else?


I'd bump it to 60s for safety, as it can take sometime for the Daemon to decrypt and compute results since we're not caching crypto results today.

jelmersnoeck:
@ianlivingstone bumped the Client Timeout to 1 minute.

jbowes:
with #207 merged, can you rebase this on master? the actual diff looks good though.

Issue #215, opened 2017-03-15
Windows: investigate setting up correct Security Attributes.


jelmersnoeck:
As mentioned in https://github.com/manifoldco/torus-cli/pull/207/commits/5b232d73319a83bb0355add2186909fb3cc81f93, we're using the default security group settings when creating a Named Pipe.

This means that we grant read access to members of the Everyone Group and the Anonymous user.

We should have a look at what exactly the impact of this is and if we can restrict that more by setting up a correct Security Attribute when creating the Named Pipe (this would require some changes to the npipe package).

ianlivingstone:
Closed by #333

PR #216, opened 2017-03-15
Readme: add Windows Named Pipe security note.

jelmersnoeck:
This adds a brief explanation about the Windows Security Attributes for
the Named Pipe we create.

As mentioned in https://github.com/manifoldco/torus-cli/issues/215, we
should investigate if we can set up stricter rules.

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,6 +63,12 @@ sudo tee /etc/apt/sources.list.d/torus.list <<< "deb https://get.torus.sh/$DISTR
   - **Environment Variables**
   - Edit **Path** in **System Variables** and add the full path to the folder where your `torus.exe` file is
 
+#### Security note
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -63,6 +63,17 @@ sudo tee /etc/apt/sources.list.d/torus.list <<< "deb https://get.torus.sh/$DISTR
   - **Environment Variables**
   - Edit **Path** in **System Variables** and add the full path to the folder where your `torus.exe` file is
 
+#### Security note
+
+Currently on Windows, when we create the named pipe, we don't set our own security attributes. This means
```
ianlivingstone: Currently on Windows, the Daemon will create a named pipe using the default security attributes. This means, that the LocalSystem account, administrators, and the creator will be granted full control. All members of the Everyone group and the anonymous account are granted read access.

More information can be found [here]..

Issue #219, opened 2017-04-01
Discussion - Support KEY=VALUE on Torus Set


ianlivingstone:
Often times I find myself typing KEY=VALUE via `torus set`, it'd be nice if we supported that syntax and split on the `=`.

`=` is not a valid name label so we could detect if a `=` is present and assume. It may or may not be surprising to the user.

Thoughts?

PR #234, opened 2017-05-21
Support `torus set KEY=value` syntax

luizbranco:
Hey, my first instinct while use `torus set` is to use the export variable syntax `KEY=value` instead of `KEY value`. This PR allows that when only one argument is passed.

Some scenarios are described in the test cases. Let me know if there is something else that needs to be covered.

Closes https://github.com/manifoldco/torus-cli/issues/219


ianlivingstone:
This is awesome, thanks so much :)

PR #235, opened 2017-05-22
Generic listener co-proceeses

enmand:
Generalize the code for creating a deamon listener co-process to keep things DRY for the Gatekeeper listener.

Mostly move functions from daemon.go to listener.go, added a Listener() interface for running and stopping a Listener process.

### DIFF
--- a/cmd/daemon.go
+++ b/cmd/daemon.go
@@ -1,25 +1,9 @@
 package cmd
 
 import (
-	"context"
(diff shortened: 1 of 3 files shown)


enmand:
Closing in favour of #242. There's no need to make the co-process spawn more generic, since only the daemon is co-process, while the Gatekeeper runs in the main Torus process.

PR #236, opened 2017-05-22
Gatekeeper

enmand:
This pull requests add a Gatekeeper Listener co-process that is able to bootstrap machine registration without manual intervention and prompting. This is useful in a continuous integration environment where servers may scale in or out automatically.

This pull requests contains code for bootstrapping AWS machines, based on their Instance Metadata and their [Instance Identity Docs](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/instance-identity-documents.html). AWS  provides signed metadata information, which can be used for instance verification.

The Gatekeeper listens on "0.0.0.0:8200" by default.

## Gatekeeper Commands
`torus gatekeeper start` -- start the Gatekeeper listener
`torus gatekeeper status` -- get the status of the Gatekeeper listener
`torus gatekeeper stop` -- stops the Gatekeeper listener

## Bootstrapping Commands:
`torus bootstrap` -- bootstrap a machine into the Gatekeeper listener.

**The `torus bootstrap` command must be run as root, since it writes to `/etc/torus`**

`bootstrap` takes additional arguments:
`--url` (_required_): Gatekeeper URL
`--auth` (_requied_): Gatekeeper authentication type to use (valid: _aws_)
`--org` (_optional_): Organization to bootstrap to
`--role` (_optional_): Team/Role to bootstrap to
`--machine` (_optional_): Machine name to provision

# Future Work

There is an opportunity for future work here, particularly around other cloud providers. For AWS, some additional verification that the server is running, and a comparison on instance running in a provided AWS environment would help Torus ensure the server actually exists. Additionally, SSL on the Gatekeeper may be worthwhile since it passes a machine secret to the server.

# Added Dependency
Added the dependency `github.com/fullsailor/pkcs7` which is used for signature validation from AWS.

# Issue Dependency
This pull request relies on #235


enmand:
Closing in favour of #242.

PR #239, opened 2017-05-23
Rename "Name" to "Full Name"

### LINKED ISSUE #233: Rename "Name" to "Display Name" on Signup
Throughout Torus there are two concepts, the name and username however it's not clear how they're different.

Let's rename `Name` to `Display Name`, so it's their purpose is most distinct!

ianlivingstone:
Throughout, we were referencing both a "Name" and a "Username", with
this change we now differentiate by asking for a "Full Name" and a
"Username".

Closes #233

### DIFF
--- a/cmd/debug.go
+++ b/cmd/debug.go
@@ -108,8 +108,8 @@ func debugInfoCmd(ctx *cli.Context) error {
 			fmt.Fprintf(w, "%s\t%s\n", "Machine Name", session.Username())
 		} else {
 			fmt.Fprintf(w, "%s\t%s\n", "User ID", session.ID())
-			fmt.Fprintf(w, "%s\t%s <%s>\n", "Identity", session.Name(), session.Email())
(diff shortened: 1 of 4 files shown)


cmd/debug.go
```diff
@@ -108,8 +108,8 @@ func debugInfoCmd(ctx *cli.Context) error {
 			fmt.Fprintf(w, "%s\t%s\n", "Machine Name", session.Username())
 		} else {
 			fmt.Fprintf(w, "%s\t%s\n", "User ID", session.ID())
-			fmt.Fprintf(w, "%s\t%s <%s>\n", "Identity", session.Name(), session.Email())
-			fmt.Fprintf(w, "%s\t%s\n", "Username", session.Username())
+			fmt.Fprintf(w, "%s\t%s <%s>\n", "Fullname <Email>", session.Name(), session.Email())
```
jbowes: missing a space

ianlivingstone: where?

PR #241, opened 2017-05-23
Fix set bug

ianlivingstone:


primitive/zz_generated_primitive.go
```diff
@@ -62,13 +62,13 @@ func (t *BaseKeyring) MarshalJSON() ([]byte, error) {
 	return b, nil
 }
 
-// Type returns the enumerated byte representation of User.
-func (t *User) Type() byte {
```
jbowes: Wanna remove this file from the PR (and open an issue to fix the flapping)?

ianlivingstone: Absolutely!

ianlivingstone: #243

PR #242, opened 2017-05-24
Gatekeeper (single process)

enmand:
This is the same implementation as #236, except it runs in the main Torus process, instead of as a co-process.

---

This pull requests add a Gatekeeper implementation that is able to bootstrap machine registration without manual intervention and prompting. This is useful in a continuous integration environment where servers may scale in or out automatically.

This pull requests contains code for bootstrapping AWS machines, based on their Instance Metadata and their [Instance Identity Docs](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/instance-identity-documents.html). AWS  provides signed metadata information, which can be used for instance verification.

The Gatekeeper listens on "0.0.0.0:8200" by default.

## Gatekeeper Commands
`torus gatekeeper start` -- start the Gatekeeper listener

## Bootstrapping Commands:
`torus machines bootstrap` -- bootstrap a machine into the Gatekeeper listener.

**The `torus machines bootstrap` command must be run as root, since it writes to `/etc/torus`**

`bootstrap` takes additional arguments:
`--url` (_required_): Gatekeeper URL
`--auth` (_requied_): Gatekeeper authentication type to use (valid: _aws_)
`--org` (_optional_): Organization to bootstrap to
`--role` (_optional_): Team/Role to bootstrap to
`--machine` (_optional_): Machine name to provision

# Future Work

There is an opportunity for future work here, particularly around other cloud providers. For AWS, some additional verification that the server is running, and a comparison on instance running in a provided AWS environment would help Torus ensure the server actually exists. Additionally, SSL on the Gatekeeper may be worthwhile since it passes a machine secret to the server.

# Added Dependency
Added the dependency `github.com/fullsailor/pkcs7` which is used for signature validation from AWS.


gatekeeper/routes/boostrap.go
```diff
@@ -0,0 +1,175 @@
...
+		machine, tokenSecret, err := api.Machines.Create(ctx, org.ID, team.ID, req.Machine.Name, nil)
+		if err != nil {
+			log.Printf("Unable to create machine: %s", err)
+			writeError(w, err)
+			return
+		}
+
+		enc := json.NewEncoder(w)
+
+		w.WriteHeader(http.StatusOK)
```
jbowes: should be a 201?

gatekeeper/bootstrap/aws/aws.go
```diff
@@ -0,0 +1,204 @@
...
+	sigData.Certificates = []*x509.Certificate{awsCert}
+
+	if err := sigData.Verify(); err != nil {
+		return fmt.Errorf("failed to verify")
+	}
+
+	var givenIdentityDoc identityDocument
+	json.Unmarshal(aws.Identity, &givenIdentityDoc)
+	var signedIdentityDoc identityDocument
+	json.Unmarshal(sigData.Content, &signedIdentityDoc)
```
jbowes: is it only the instanceId and pendingTime that's signed, or is it the full document? Couldn't you just compare the raw bytes of the identity document with the pkcs#7 content?

gatekeeper/bootstrap/aws/aws.go
```diff
@@ -0,0 +1,204 @@
+// Package aws provides Gatekeeper authentication for AWS
+package aws
```
jbowes: This file contains logic for both  the client side and the server side. Its not immediately clear to me when reading this where each responsibility lies. Maybe that separation could be clarified?

gatekeeper/client/client.go
```diff
@@ -0,0 +1,67 @@
...
+	var bootresp apitypes.BootstrapResponse
+	resp, err := c.rt.Do(context.Background(), req, &bootresp)
+	if err != nil {
+		return nil, err
+	}
+	defer resp.Body.Close()
+
+	if bootresp.Error != "" {
+		return nil, errors.New(bootresp.Error)
+	}
```
jbowes: Will this ever happen? if the response code is not 2XX, the RequestDoer will attempt to populate apitypes.Error:
https://github.com/manifoldco/torus-cli/blob/master/registry/round_tripper.go#L107

gatekeeper/routes/routes.go
```diff
@@ -0,0 +1,24 @@
...
+	return mux
+}
+
+func newMachineBootstrapRoutes(org, team string, api *api.Client) *bone.Mux {
+	mux := bone.New()
+
+	mux.PostFunc("/aws", awsBootstrapRoute(org, team, api))
+
+	return mux
+}
```
jbowes: This file is a lot of boilerplate for a single endpoint; how about just wiring it up in the Listen func? when there are new auth types added, we can rework the route declaration then.

cmd/bootstrap.go
```diff
@@ -0,0 +1,87 @@
...
+	bootstrap := cli.Command{
+		Name:     "bootstrap",
+		Usage:    "Bootstrap a new machine using Torus Gatekeeper",
+		Category: "SYSTEM",
+		Flags: []cli.Flag{
+			authProviderFlag("Auth provider for bootstrapping", true),
+			machineFlag("Machine name to bootstrap", false),
+			urlFlag("Gatekeeper URL for bootstrapping", true),
+			orgFlag("Org the machine will belong to", false),
+			roleFlag("Role the machine will belong to", false),
```
jbowes: It would be cool to think about how the machine name, org, and role could be handled by the auth mechanism itself, on the gatekeeper side. Not for this pr though, of course :)

gatekeeper/routes/boostrap.go
```diff
@@ -0,0 +1,175 @@
...
+		if err := provider.Verify(); err != nil {
+			log.Printf("Instance verification failed: %s", err)
+			writeError(w, fmt.Errorf("instance verification failed: %s", err))
+		}
+
+		if req.Machine.Org != "" {
+			orgName = req.Machine.Org
+		}
+		if orgName == "" {
+			log.Printf("No organization provided to bootstrap: %s", err)
```
jbowes: err will be nil here.

gatekeeper/routes/boostrap.go
```diff
@@ -0,0 +1,175 @@
...
+			return
+		}
+
+		provider := aws.New()
+		provider.Identity = req.Identity
+		provider.Signature = req.Signature
+
+		if err := provider.Verify(); err != nil {
+			log.Printf("Instance verification failed: %s", err)
+			writeError(w, fmt.Errorf("instance verification failed: %s", err))
```
jbowes: missing a return

cmd/bootstrap.go
```diff
@@ -0,0 +1,87 @@
...
+func init() {
+	bootstrap := cli.Command{
+		Name:     "bootstrap",
+		Usage:    "Bootstrap a new machine using Torus Gatekeeper",
+		Category: "SYSTEM",
+		Flags: []cli.Flag{
+			authProviderFlag("Auth provider for bootstrapping", true),
+			machineFlag("Machine name to bootstrap", false),
+			urlFlag("Gatekeeper URL for bootstrapping", true),
+			orgFlag("Org the machine will belong to", false),
```
jbowes: this is required, afaict

cmd/bootstrap.go
enmand: I was thinking about that a bit... it might be possible using IAM, or reading tags from the given EC2 instance maybe?

gatekeeper/bootstrap/aws/aws.go
enmand: I think it's probably safe to remove the logic on the client side. It does a pre-flight validation on the instance signature before it sends it to the Gatekeeper, but that is probably unneeded since the Gatekeeper does this check anyway.

gatekeeper/client/client.go
```diff
@@ -0,0 +1,67 @@
...
+	"github.com/manifoldco/torus-cli/registry"
+)
+
+const (
+	gatekeeperAPIVersion = "v0"
+	clientTimeout        = time.Minute
+)
+
+type clientRoundTripper struct {
+	registry.DefaultRequestDoer // TODO: Could be a generic RequestDoer?
```
enmand: I was thinking instead of using the same `registry.DefaultRequestDoer` it would make sense to move that to a generic package, since this client isn't talking to the registry. I'll update the comment in the code too

gatekeeper/http/http.go
```diff
@@ -0,0 +1,85 @@
...
+	return g.hd.Stop()
+}
+
+// Addr returns the address of the running Gatekeeper service
+func (g *Gatekeeper) Addr() string {
+	return g.s.Addr
+}
+
+// loggingHandler logs requests in the format: METHOD URL
+func loggingHandler(next http.Handler) http.Handler {
```
enmand: Definitely a dupe -- would it make more sense to add an `http` package with a `middleware` that contains some common HTTP middleware, or would that be too much?

gatekeeper/gatekeeper.go
```diff
@@ -0,0 +1,60 @@
...
+	"github.com/urfave/cli"
+
+	"github.com/manifoldco/torus-cli/api"
+	"github.com/manifoldco/torus-cli/config"
+	"github.com/manifoldco/torus-cli/gatekeeper/http"
+)
+
+// Gatekeeper provides the Listener interface for a daemonized HTTP co-process that will act
+// to register machines
+type Gatekeeper struct {
```
enmand: It was used before in the co-process more since it had the `Listener` interface implemented, but now that it's a single process, it's not necessary.

gatekeeper/bootstrap/aws/bootstrap.go
```diff
@@ -0,0 +1,102 @@
...
+	defer resp.Body.Close()
+
+	b, err := ioutil.ReadAll(resp.Body)
+	if err != nil {
+		return nil, err
+	}
+	return b, nil
+}
+
+type identityDocument struct {
```
enmand: This is used by `Verifier` too, so I think there is a better place for it

gatekeeper/bootstrap/bootstrap.go
```diff
@@ -0,0 +1,29 @@
...
+	"github.com/manifoldco/torus-cli/gatekeeper/apitypes"
+	"github.com/manifoldco/torus-cli/gatekeeper/bootstrap/aws"
+)
+
+const (
+	// AWSPublic is Amazon's Public Cloud Provider
+	AWSPublic = "aws"
+)
+
+// Handler the function returned by the Bootstrap server that can
```
enmand: I don't know how I feel about this type, and returning it from `bootstrap.New` directly. Thoughts?

cmd/gatekeeper.go
```diff
@@ -0,0 +1,67 @@
...
+	gatekeeper := cli.Command{
+		Name:     "gatekeeper",
+		Usage:    "Manage the machine gatekeeper",
+		Category: "SYSTEM",
+		Subcommands: []cli.Command{
+			{
+				Name:  "start",
+				Usage: "Start the machine gatekeeper",
+				Flags: []cli.Flag{
+					cli.BoolFlag{
```
enmand: I don't think this is needed anymore. Could I do `config.Load()` below instead since the daemon would already be started?

cmd/gatekeeper.go
```diff
@@ -0,0 +1,67 @@
...
+		},
+	}
+
+	Cmds = append(Cmds, gatekeeper)
+}
+
+// startGatekeeper starts the machine Gatekeeper
+func startGatekeeperCmd(ctx *cli.Context) error {
+	noPermissionCheck := ctx.Bool("no-permission-check")
+	torusRoot, err := config.CreateTorusRoot(!noPermissionCheck)
```
enmand: I think this should be `cfg, err := config.Load()` which can get rid of a lot below, since the Daemon is already running

gatekeeper/apitypes/bootstrap.go
```diff
@@ -0,0 +1,32 @@
+package apitypes
```
enmand: I wasn't sure if this should go in it's own package under `gatekeeper`, or added to the `apitypes` base package. Thoughts?

gatekeeper/bootstrap/bootstrap.go
```diff
@@ -0,0 +1,29 @@
+// Package bootstrap provides authentication actions for the bootstrap process
```
enmand: This whole file might be overkill until more ways to bootstrap are created.

cmd/bootstrap.go
enmand: I think either it should be required, or both it and `role` are not required. The Gatekeeper should try and load it's defaults and use that for machine bootstrapping, but that might not be a predictable way to determine the machine's organization or team

gatekeeper/http/http.go
jbowes: I think it makes sense, but could be done in another PR, at another time. A comment here and an issue to track it would be awesome.

gatekeeper/routes/boostrap.go
```diff
@@ -0,0 +1,174 @@
...
+			Secret: tokenSecret,
+		}
+
+		enc.Encode(resp)
+	}
+}
+
+// writeError returns a bootstrapping error
+func writeError(w http.ResponseWriter, err error) {
+	w.WriteHeader(http.StatusInternalServerError)
```
jbowes: what about different status codes for the requests? should a bad cert return a diff code than an empty org name?

gatekeeper/bootstrap/aws/verifier.go
```diff
@@ -0,0 +1,103 @@
...
+		return nil
+	}
+
+	sigData.Certificates = []*x509.Certificate{awsCert}
+
+	if err := sigData.Verify(); err != nil {
+		return fmt.Errorf("failed to verify")
+	}
+
+	if !bytes.Equal(v.Identity, sigData.Content) {
```
jbowes: This is subject to a timing attack

gatekeeper/apitypes/bootstrap.go
```diff
@@ -0,0 +1,32 @@
...
+	Name string `json:"name"`
+	Org  string `json:"org_id"`
+	Team string `json:"team_id"`
+}
+
+// BootstrapResponse represents the Response object returned to Gatekeeper bootstrap request
+type BootstrapResponse struct {
+	Token  *identity.ID  `json:"token"`
+	Secret *base64.Value `json:"secret"`
+	Error  string        `json:"error"`
```
jbowes: the Error field is unused now IIUC

cmd/gatekeeper.go
jbowes: agreed

cmd/gatekeeper.go
jbowes: yup, that makes sense to me! maybe just add a comment to that effect; that the gatekeeper needs the root dir, but assumes the daemon has already created it.

gatekeeper/bootstrap/bootstrap.go
jbowes: since it isn't reused, the external signature could replace both `New` and `Handler` with `bootstrap.Do(t, url, name, org, role string)`

If the handler were going to be called again, returning it would make sense.

ianlivingstone:
Before we merge this up, can we get an issue (or many?) created for tracking what we'd need to do to take this from a prototype into a beta form.  For example, supporting role retrieval based on aws instance tags, documentation, getting started guide, etc.

Looks great 💯

jbowes:
Good to go! Can you rebase on master, and squash down the commits, and then we can merge it.

PR #244, opened 2017-05-27
policies test cmd initial implementation

bertoldia:
This MR is a first attempt at the `torus policies test` command as specified in Ian's email.

Testing a path to see if all relevant policy allows access to it goes something like this:
- validate input
- fetch org
- fetch user teams
    - fetch user profile
    - fetch all teams for org
    - fetch memberships for each team and filter team by user membership
- fetch all policies for org
- fetch all policy attachments for org
- filter policies
    - by team attachment
    - by path
    - by action
- test access (yes if >= 1 policies allow access and 0 deny access, false otherwise).

Notes:
- I've followed the existing pattern in policies.go and elsewhere in other command implementations that making API calls can be  expensive and should be done in parallel when possible.
- When implementing the above pattern, I used the existing idiom of go routine + closure + WaitGroups. Given more time I might have tried an approached that used channels instead, and seen how the two compared.
- There is a complete lack of automated testing here, unit, integration or otherwise. I'm usually pretty good about writing test (really I am), but given the time scope I skipped this.
- Testing was necessarily manual, and I probably missed some scenarios so my implementation may have logic errors.
- There is some code and idea duplication, both introduced by me and to a lesser extent already existing. Ideally I'd refactor/consolidate to eliminate this, but given the scope of the task I did not take this on.
- There is one part of the implementation I'm pretty sure is wrong; the whole path parsing and comparison bit. For starters the pathexp.PathExp class does not represent secrets so I totally ignored secrets in the implementation  (as is done in cmd.allow.go). I suspect this part is not quite right. Also, I'm pretty sure comparing the specified path with a Policy's Resources is not always right. Fixing this should be simple though for someone more familiar with path expression.

Enjoy!

### DIFF
--- a/cmd/policies.go
+++ b/cmd/policies.go
@@ -480,3 +480,32 @@ func filterTeamsByUser(client *api.Client, teams []envelope.Team,
 		close(teamChan)
 	}()
 }
+
+// Fetch all policies and policy-attachments for the org. The two steps are
+// independent so do them in parallel.
+func getPoliciesAndAttachments(client *api.Client, orgID *identity.ID) ([]envelope.Policy, []envelope.PolicyAttachment, error) {
+	c := context.Background()
+	wg := &sync.WaitGroup{}
+	wg.Add(2)
+
+	var pErr error
+	var policies []envelope.Policy
+	go func() {
+		defer wg.Done()
+		policies, pErr = client.Policies.List(c, orgID, "")
+	}()
+
+	var aErr error
+	var attachments []envelope.PolicyAttachment
+	go func() {
+		defer wg.Done()
+		attachments, aErr = client.Policies.AttachmentsList(c, orgID, nil, nil)
+	}()
+	wg.Wait()
+
+	if aErr != nil || pErr != nil {
+		return nil, nil, cli.NewMultiError(pErr, aErr,
+			errs.NewExitError(policyTestFailed))
+	}
+	return policies, attachments, nil
+}
--- a/cmd/policies.go
+++ b/cmd/policies.go
@@ -580,3 +580,16 @@ func policyTouchesPathPredicate(pathExp *pathexp.PathExp) PolicyPredicate {
 		return false
 	}
 }
+
+// Create a predicate to filter (include) policies that implement the specified
+// policy-action ([crudl]) in any one of their statements.
+func policyImplementsActionPredicate(action primitive.PolicyAction) PolicyPredicate {
+	return func(policy envelope.Policy) bool {
+		for _, s := range policy.Body.Policy.Statements {
+			if s.Action&action > 0 {
+				return true
+			}
+		}
+		return false
+	}
+}
--- a/cmd/policies.go
+++ b/cmd/policies.go
@@ -559,3 +559,24 @@ func policyAttachedToTeamsPredicate(teams []envelope.Team,
 		return false
 	}
 }
+
+// Create a predicate to filter (exclude) policies that do not apply to the
+// specified path. A policy applies to a path if any of its resources (paths)
+// are equivalent to the specified path. The secret portion of the path is
+// ignored.
+func policyTouchesPathPredicate(pathExp *pathexp.PathExp) PolicyPredicate {
+	return func(policy envelope.Policy) bool {
+		for _, s := range policy.Body.Policy.Statements {
+			rp, err := parseRawPath(s.Resource)
+			if err != nil {
+				// TODO Log this
+				continue
+			}
+			// I'm pretty sure this is wrong...
+			if rp.Equal(pathExp) {
+				return true
+			}
+		}
+		return false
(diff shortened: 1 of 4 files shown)


cmd/policies.go
```diff
@@ -309,3 +325,345 @@ func viewPolicyCmd(ctx *cli.Context) error {
...
+	path, err := parseRawPath(rawPath)
+	if err != nil {
+		return nil, nil, nil, errs.NewErrorExitError(policyTestFailed, err)
+	}
+	return &action, &userName, path, nil
+}
+
+// Parse and validate a raw path, possibly including a secret. The secret
+// portion is discarded.
+func parseRawPath(rawPath string) (*pathexp.PathExp, error) {
```
jbowes: could this func be shared with the code in cmd/allow?

cmd/policies.go
```diff
@@ -309,3 +325,345 @@ func viewPolicyCmd(ctx *cli.Context) error {
...
+
+	if aErr != nil || pErr != nil {
+		return nil, nil, cli.NewMultiError(pErr, aErr,
+			errs.NewExitError(policyTestFailed))
+	}
+	return policies, attachments, nil
+}
+
+// Test if a Policy satisfies some condition
+type PolicyPredicate func(envelope.Policy) bool
```
jbowes: The predicates/allpredicate is super nice

cmd/policies.go
```diff
@@ -309,3 +325,345 @@ func viewPolicyCmd(ctx *cli.Context) error {
...
+			if isMachineTeam(t.Body) {
+				continue
+			}
+
+			wg.Add(1)
+			go func(team envelope.Team) {
+				defer wg.Done()
+				members, err := client.Memberships.List(c, orgID, team.ID, nil)
+				if err != nil {
+					// TODO: Log this!!!
```
jbowes: can this TODO be done in the pr?

cmd/policies.go
```diff
@@ -309,3 +325,345 @@ func viewPolicyCmd(ctx *cli.Context) error {
...
+		return err
+	}
+
+	action, userName, path, err := parseArgs(ctx)
+	if err != nil {
+		return err
+	}
+
+	client := api.NewClient(cfg)
+	c := context.Background()
```
jbowes: This context should be shared down through all the funcs; their usage should be hierarchical (though in this case, just one master one will do).

cmd/policies.go
```diff
@@ -309,3 +325,345 @@ func viewPolicyCmd(ctx *cli.Context) error {
...
+			if !pred(policy) {
+				return false
+			}
+		}
+		return true
+	}
+}
+
+func filterPolicies(policies []envelope.Policy, predicate PolicyPredicate) []envelope.Policy {
+	result := make([]envelope.Policy, 0)
```
jbowes: It's not necessary to`make` a slice; you can start with the zero value/nil and just append to that.

cmd/policies.go
```diff
@@ -309,3 +325,345 @@ func viewPolicyCmd(ctx *cli.Context) error {
...
+	for _, pol := range policies {
+		if predicate(pol) {
+			result = append(result, pol)
+		}
+	}
+	return result
+}
+
+// Crete a Predicate to filter (exclude) policies that are not attached to any
```
jbowes: godoc comments always start with the func name, like:
` policyAttachedToTeamsPredicate creates a Predicate to...`

cmd/policies.go
```diff
@@ -309,3 +325,345 @@ func viewPolicyCmd(ctx *cli.Context) error {
...
+		return false
+	}
+}
+
+// Given a set of policies, determine if they allow access (i.e. allow a
+// specific action on a specific path). Returns true if at least one policy
+// allows access AND exactly zero policies deny access. Return false otherwise.
+// This method only make sense if the policies have been previously reduced to
+// those referring to the same resource and action.
+func policiesAllowAccess(policies []envelope.Policy) bool {
```
jbowes: very nice.

cmd/policies.go
```diff
@@ -309,3 +325,345 @@ func viewPolicyCmd(ctx *cli.Context) error {
...
+// ignored.
+func policyTouchesPathPredicate(pathExp *pathexp.PathExp) PolicyPredicate {
+	return func(policy envelope.Policy) bool {
+		for _, s := range policy.Body.Policy.Statements {
+			rp, err := parseRawPath(s.Resource)
+			if err != nil {
+				// TODO Log this
+				continue
+			}
+			// I'm pretty sure this is wrong...
```
jbowes: It is, but hey, you know it! :)
The pathexp pkg would need to add a Contains method to the PathExp object. That's a decent sized chunk of work, so Equality is a-ok for now.

jbowes:
Thanks!

PR #256, opened 2017-06-01
Ensure GITHUB_TOKEN is set before releasing

ianlivingstone:
If a `GITHUB_TOKEN` was not set, you could still release to production
but updating the GitHub release would fail.

To fix this, I've added a check to ensure a `GITHUB_TOKEN` is set before
continuing.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -172,6 +172,11 @@ ifneq (v$(VERSION),$(shell git describe --tags --dirty))
 	$(error "VERSION $(VERSION) is not git HEAD")
 endif
 
+githubcheck:
+ifeq (,$(GITHUB_TOKEN))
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -253,7 +258,7 @@ $(addprefix deb-,$(LINUX)): deb-%: binary-% builds/deb deb-container
 		cp /torus/contrib/systemd/torus.service \
 			deb-tmp/torus/lib/systemd/system && \
 		cp /torus/contrib/systemd/token.environment \
-			deb-tmp/torus/etc/torus && \
+			eeb-tmp/torus/etc/torus && \
```
jbowes: eeb?

ianlivingstone: I'll fix

PR #260, opened 2017-06-10
Allow set cmd to read from a file piped to stdin

### LINKED ISSUE #218: Support `-f` on torus set
Many times, secrets are stored in a file, it'd be nice if you could import secrets from a file into torus using the set command.

Something like:

```
$ torus set -e myprod -f secrets.yml
```

Where the supported files are yaml, json, or env

luizbranco:
If no arguments are passed to `set` we try to check if there is any input being piped to stdin. Closes #218 

Given a `file.txt` like:

```
user=root
password=s3cr3t
port=3000
```

We can now do:

```shell
torus set --org example --project mvp < file.txt
```

It is also useful to export all secrets from one organisation / project and add to another one:

```
torus view --org example --project foo > secrets.txt
torus set  --org another --project bar < secrets.txt
```

@ianlivingstone I'm currently doing a round trip to set each secret. I'm not sure if there is a way to send multiple secrets in a single payload.

### DIFF
--- a/cmd/set.go
+++ b/cmd/set.go
@@ -2,6 +2,7 @@ package cmd
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"strings"
 
@@ -45,9 +46,11 @@ func init() {
 }
 
 func setCmd(ctx *cli.Context) error {
-	key, value, errMsg := parseSetArgs(ctx.Args())
-	if errMsg != "" {
-		return errs.NewUsageExitError(errMsg, ctx)
+	args := ctx.Args()
+	key, value, err := parseSetArgs(args)
+
+	if err != nil {
+		return errs.NewUsageExitError(err.Error(), ctx)
 	}
 
 	cred, err := setCredential(ctx, key, func() *apitypes.CredentialValue {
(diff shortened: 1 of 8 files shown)


ianlivingstone:
> @ianlivingstone I'm currently doing a round trip to set each secret. I'm not sure if there is a way to send multiple secrets in a single payload.

It'd be non-trivial to add `multi-set` support (it'd require modifying the Daemon Logic Engine to support an array of Credentials instead of just one). All of the backing bits on the server support it already though.

So perhaps, we get this merged in (once we've reviewed it), and open an issue to handle this as a single payload?

cmd/set.go
```diff
@@ -45,48 +53,126 @@ func init() {
...
+	var lines []string
+
+	scanner := bufio.NewScanner(os.Stdin)
+	for scanner.Scan() {
+		line := scanner.Text()
+
+		if line != "" {
+			lines = append(lines, line)
+		}
+	}
```
jbowes: It would be very cool if you incorporated something like https://github.com/google/shlex into this, treating the contents as a shell style .env file. Then comments and quoting/escaping would be handled in the same way as other tools handle env configuration in files.

cmd/set.go
```diff
@@ -45,48 +53,126 @@ func init() {
...
-func parseSetArgs(args []string) (key string, value string, errMsg string) {
-	// If there is only one argument, try to parse using env var syntax: KEY=value
+// parseSetSecrets returns a list of secret pairs either from the command line
+// arguments or from the standard input. It returns an error if there is a
+// problem parsing secrets or stdin fails to read.
+func parseSetSecrets(args []string) ([]secretPair, error) {
+	if len(args) > 0 {
+		secret, err := parseSetArgs(args)
+		return []secretPair{secret}, err
+	}
```
jbowes: I'm concerned with falling back to stdin when no args are given; someone using torus for the first time and running `torus set` with no args may just think it has hung.

Instead, how about adding a `--env` (or similar) argument, which accepts either a file name to read, or the conventional `-` as file name to indicate stdin?

jeffandersen: Would it make sense to introduce this behavior as a separate command?

ie. `torus import [filename]` (or naked to accept stdin)

jbowes: I think so!

luizbranco: Added as a separated command. I'm reusing the same `setCredential` function from the set cmd and flags. Let me know if that's ok or if they should be in a common file.

cmd/set.go
luizbranco: Makes sense. I will try to incorporate that during the weekend.

ianlivingstone:
This is awesome :)

luizbranco:
@jbowes added google/shlex to the `import` command. Can you please take a second look?

There is a small incompatibility with the `view` command: when multiple words are used for the value, the output is not surround by quotes:

```
BAR=bar
BAZ=baz word
```

I'm going to create an issue for that.

jbowes:
Awesome! Thank you :)

Issue #248, opened 2017-05-27
Gatekeeper: Bootstrap over SSL


enmand:
Currently, all bootstrapping communication is done over a normal HTTP connection, however machine token ID and secret are shared back to the machine that is being bootstrapped. An evesdropper in your network could MITM to retrieve this information, and once the machine was given access, a malicious user could act as that machine.

PR #262, opened 2017-06-14
Add optional SSL for Gatekeeper

enmand:
This pull request starts to add optional SSL for the Gatekeeper bootstrap tool. This would start progress on #248.

This adds --cert, --key (`TORUS_GATEKEEPER_CERT` and `TORUS_GATEKEEPER_CERT_KEY`) to the `torus gatekeeper start` command, and `--ca` (`TORUS_BOOTSTRAP_CA`) to the `torus machines bootstrap` to start the service with TLS.

Some considerations: the certificate would need to be signed by a CA in the bootstrapping system's CA bundle. There may be cases where a user would use a self-signed cert, or a cert signed by an internal CA. In that case, the certificates would need to be added to the system's certificate bundle, or else the bootstrapping will fail on a bad certificate. The `--ca` flag will override using the system's certificate bundle.

It might be something to explore to do some automatic TLS through something like LetsEncrypt, if no certificates are provided, so that Gatekeeper is always over SSL.

### DIFF
--- a/gatekeeper/client/client.go
+++ b/gatekeeper/client/client.go
@@ -3,7 +3,11 @@ package client
 
 import (
 	"context"
+	"crypto/tls"
+	"crypto/x509"
 	"fmt"
+	"io/ioutil"
+	"log"
 	"net/http"
 	"strings"
 	"time"
@@ -29,20 +33,47 @@ type Client struct {
 }
 
 // NewClient returns a new client to a Gatekeeper host that can bootstrap this machine
-func NewClient(host string) *Client {
+func NewClient(host, caFile string) (*Client, error) {
 	if !strings.HasSuffix(host, "/") {
 		host += "/"
 	}
+
+	caPool, err := x509.SystemCertPool()
+	if err != nil {
+		log.Printf("Could not load system certificate pool: %s. Creating custom pool", err)
+		caPool = x509.NewCertPool()
+	}
+
+	if caFile != "" {
+		caBytes, err := ioutil.ReadFile(caFile)
+		if err != nil {
+			return nil, fmt.Errorf("unable to read ca file: %s", err)
+		}
+		ok := caPool.AppendCertsFromPEM(caBytes)
+		if !ok {
+			return nil, fmt.Errorf("unable to parse and append ca file: %s", err)
+		}
+	}
+
+	tlsConfig := &tls.Config{
+		RootCAs: caPool,
+	}
+
+	transport := &http.Transport{
+		TLSClientConfig: tlsConfig,
+	}
+
 	return &Client{
 		rt: &clientRoundTripper{
 			DefaultRequestDoer: registry.DefaultRequestDoer{
 				Client: &http.Client{
-					Timeout: clientTimeout,
+					Timeout:   clientTimeout,
+					Transport: transport,
 				},
 				Host: host,
 			},
 		},
-	}
+	}, nil
 }
 
 // Bootstrap bootstraps the machine with Gatekeeper
--- a/gatekeeper/http/http.go
+++ b/gatekeeper/http/http.go
@@ -1,6 +1,9 @@
 package http
 
 import (
+	"crypto/tls"
+	"fmt"
+	"io/ioutil"
 	"log"
 	"net/http"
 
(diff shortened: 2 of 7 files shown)


ianlivingstone:
> Some considerations: the certificate would need to be signed by a CA in the bootstrapping system's CA bundle. There may be cases where a user would use a self-signed cert, or a cert signed by an internal CA. In that case, the certificates would need to be added to the system's certificate bundle, or else the bootstrapping will fail on a bad certificate.

Interesting, we could also just offer it as a configuration option via the `.torusrc` file/env/flag to include a self signed CA along with the bundle. 

> It might be something to explore to do some automatic TLS through something like LetsEncrypt, if no certificates are provided, so that Gatekeeper is always over SSL.

I hadn't thought of this, it'd definitely be neat, I wonder if there would be a good way for us to do it through some sort of plugin or external tool.

gatekeeper/http/http.go
```diff
@@ -28,11 +31,35 @@ type Gatekeeper struct {
 }
 
 // NewGatekeeper returns a new Gatekeeper.
-func NewGatekeeper(org, team string, cfg *config.Config, api *api.Client) *Gatekeeper {
+func NewGatekeeper(org, team, certpath, keypath string, cfg *config.Config, api *api.Client) (*Gatekeeper, error) {
 	server := &http.Server{
 		Addr: cfg.GatekeeperAddress,
 	}
 
+	if certpath == "" || keypath == "" {
```
ianlivingstone: shouldn't this be an `&&` as they are both required?

gatekeeper/http/http.go
```diff
@@ -28,11 +31,35 @@ type Gatekeeper struct {
...
+		log.Printf("Starting Gatekeeper without SSL")
+	} else {
+		keypair, err := tlsKeypair(certpath, keypath)
+		if err != nil {
+			return nil, err
+		}
+
+		tlsConfig := &tls.Config{
+			MinVersion:               tls.VersionTLS12,
+			CurvePreferences:         []tls.CurveID{tls.CurveP521, tls.CurveP384, tls.CurveP256},
```
ianlivingstone: Have we checked the right configuration for these curves (e.g. are they the widely recommended ones)? If I remember correclty these curves have been backdoored by the NSA (since they're NIST curves).

gatekeeper/http/http.go
ianlivingstone: or better: have tlsKeypair error appropriately if both are not provided (or above someplace)..

gatekeeper/http/http.go
enmand: I found this configuration based on some research online, but it may be better to remove `CurvePreferences` (maybe even `CipherSuites`, I think `PreferServerCipherSuites` can be removed safely in any case) from the config. Based on https://github.com/golang/go/blob/46f4bfb2d17a3ccb4b3207d086a90cac3c00ea2f/src/crypto/tls/common.go#L688 it looks like the defaults are `[]CurveID{X25519, CurveP256, CurveP384, CurveP521}` for `CurvePreferences`. Defaults for  `CipherSuites` depend on the hardware (https://github.com/golang/go/blob/46f4bfb2d17a3ccb4b3207d086a90cac3c00ea2f/src/crypto/tls/common.go#L917)

ianlivingstone:
This looks great, can you rebase it?

enmand:
@ianlivingstone -- just added the --ca flag for bootstrapping against a custom CA. I used [square/certstrap](https://github.com/square/certstrap) to create a fake CA to test against. If that looks good, I can rebase.

Regarding something like LetsEncrypt, the more I think about it, the less sense it might make. There's a few scenarios where it would make sense (if a user doesn't have any certificates and doesn't know to/want to get certs from somewhere), but overall something like [lego](https://github.com/xenolf/lego) would be an easy way for anyone to get a LetsEncrypt certificate. There would also be some additional configuration that would need to be provided to gatekeeper (probably LetsEncrypt email address and hostname(s) for certificate) would might confuse things more. That said, the [acme](https://godoc.org/github.com/xenolf/lego/acme) package from lego simplifies things like verification, and could help with the auto-TLS stuff on Gatekeeper.

gatekeeper/client/client.go
```diff
@@ -29,20 +32,46 @@ type Client struct {
...
 
 // NewClient returns a new client to a Gatekeeper host that can bootstrap this machine
-func NewClient(host string) *Client {
+func NewClient(host, caFile string) (*Client, error) {
 	if !strings.HasSuffix(host, "/") {
 		host += "/"
 	}
+
+	caPool, err := x509.SystemCertPool()
+	if err != nil {
```
ianlivingstone: Should we log the error? As in, "Could not load System Cert Pool, error: %s, creating custom certificate pool"?

Also, thinking about this more, I think this would be fatal if a user didn't specify the `--ca` flag?

enmand: > Also, thinking about this more, I think this would be fatal if a user didn't specify the --ca flag?

Not sure here. There are a few cases I see:
1. No `--key` and `--cert` given on `gatekeeper` cmd will start the Gatekeeper without SSL. Maybe they should be required too (`--ca` would probably need them to be required to be consistent).
2. `--key` and `--cert` specify a certificate that is signed by the system bundle (or an intermediate cert), then `--ca` wouldn't be required.

Thinking on what OpenShift does, I think (1) makes sense. In OpenShift, they generate a CA cert, and then use that as a basis of the certificates in the system. That way, there's no reliance on having a certain CA be trusted or not in your system. That should force Gatekeeper over SSL in all cases (which is probably the right thing)

ianlivingstone: For testing and development purposes, it makes sense for the Gatekeeper to not require SSL. However, in all other cases SSL should be required.

So, perhaps, we just add a log to create noise (a WARN) about the fact that SSL is not enabled.

Regarding the error, in the case where `--ca` is not provided *and* we can't get the SystemCertPool, we'd end up swallowing the error. Instead, I think this should be fatal instead of silently swallowed with an empty cert pool. Perhaps I'm missing some context?

ianlivingstone:
@enmand can you rebase this branch and then we can merge it in :)

PR #266, opened 2017-06-17
Wrap secret value with multiple words in quotes using view command

### LINKED ISSUE #265: View command doesn't surround multiple words with quotes
As mentioned on #260 

Current output:
```
BAR=bar
BAZ=baz word
```

Expected output:
```
BAR=bar
BAZ="baz word"
```

luizbranco:
Closes #265

I changed the printXXXFormat APIs to accept an `io.Writer` instead of always using `os.Stdout`. Makes easier to test and we can combine with another writers if needed.

I also removed the hint from the view command because the text is added to the output file when redirecting. 
Eg: `torus view > secret.env` has at the end:

```
[1mProtip: [0mDefine an organization and project for your current working directory using `torus link`
```

Probably a better approach would be to disable hints whenever stdout is not printing to the terminal.

### DIFF
--- a/cmd/view.go
+++ b/cmd/view.go
@@ -81,10 +81,14 @@ func writeEnvFormat(w io.Writer, secrets []apitypes.CredentialEnvelope, path str
 	tw := tabwriter.NewWriter(w, 2, 0, 2, ' ', 0)
 
 	for _, secret := range secrets {
-		value := (*secret.Body).GetValue()
+		value := (*secret.Body).GetValue().String()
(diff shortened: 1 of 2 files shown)


ianlivingstone:
This is awesome!

I'm curious as to why the hints were still showing, we had disabled them in https://github.com/manifoldco/torus-cli/pull/254.

luizbranco:
I just tested again, you are correct, there is no hint when redirecting stdout. I must have done something wrong previously. I've added the hint back to `view`.

PR #307, opened 2017-11-06
Began 'orgs members' cmd

### LINKED ISSUE #301: torus orgs members - Making it easier to understand who has access to what
Today, it's really hard to get a grasp on who has access to which resources and secrets. One of the reasons is that there isn't a good way to get a list of *all* users in an org and the teams they belong to. Instead, a user has to go through each team via `torus teams members <team-name>` to get this information.

One step to towards making this better may be to introduce a `torus orgs members` command which lists all members in an organization and the teams they belong to. Like all other commands, this resolve the target org via context (e.g. linked `.torus.json` in path? flag? environment variable) in that order of priorty).

It'd be nice if you could also specify a `--team, -t` flag to show the teams that all members of the devops team belong to.  If multiple `--team, -t` flags were specified then it could be an intersection (all members who belong to all of the specified teams).

### Examples 

**Using flags to list members**

```
$ torus orgs members --org myorg
USERNAME                       NAME                        ROLE                      TEAMS 
ian                                      ian livingstone          admin                     admin, member, devops
jake                                    jake the dog              member                 member
franklin                              manny                        member                 member, devops
```

**Inside a linked directory**

```
$ torus orgs members
USERNAME                       NAME                        ROLE                      TEAMS 
ian                                      ian livingstone          admin                     admin, member, devops
jake                                    jake the dog              member                 member
```

**Using `-t` flag**

```
$ torus orgs members -t devops
USERNAME                       NAME                        ROLE                      TEAMS 
ian                                      ian livingstone          admin                     admin, member, devops
franklin                              manny                        member                 member, devops
```

**Using multiple `-t` flags**

```
$ torus orgs members -o myorg -t devops -t admin
USERNAME                       NAME                        ROLE                      TEAMS 
ian                                      ian livingstone          admin                     admin, member, devops
```

### Notes

- Teams would be listed by importance (e.g. system team first, owner, admin, then member)
- Users would be listed by their role first then alphabetically by username (e.g. owners at top, then admin, then member (everyone else))

### Questions

- Technically there is no primitive inside Torus known as a ROLE for a user, it's an implied state if you belong to an owner or admin team. Should it be called out here? Or just left to list the teams they belong to? Is it just making things more confusing or does it simplify the current system?
- How do we display which org/teams we're currently listing? Should there be a context line that says (listing members for X org with memberships to Z,Y teams)?

Open to any and all feedback :)

MattAWright:
Closes #301

### DIFF
--- a/cmd/orgs.go
+++ b/cmd/orgs.go
@@ -4,11 +4,10 @@ import (
 	"context"
 	"fmt"
 	"os"
+	"sort"
 	"strconv"
-	"strings"
 	"sync"
 	"text/tabwriter"
-	"unicode/utf8"
 
 	"github.com/urfave/cli"
 
@@ -19,7 +18,7 @@ import (
 	"github.com/manifoldco/torus-cli/errs"
 	"github.com/manifoldco/torus-cli/hints"
 	"github.com/manifoldco/torus-cli/identity"
-	"github.com/manifoldco/torus-cli/promptui"
+	"github.com/manifoldco/torus-cli/primitive"
 )
 
 func init() {
@@ -233,10 +232,30 @@ func orgsRemove(ctx *cli.Context) error {
 	return nil
 }
 
-func orgsMembersListCmd(ctx *cli.Context) error {
+// ByTeamPrecedence will implement the sort Interface
+type ByTeamPrecedence []envelope.Team
 
-	// Use "member" team name to bypass specific teams
-	teamName := "member"
+func (tp ByTeamPrecedence) Len() int {
+	return len(tp)
+}
+
+func (tp ByTeamPrecedence) Swap(i, j int) {
+	tp[i], tp[j] = tp[j], tp[i]
+}
+
+func (tp ByTeamPrecedence) Less(i, j int) bool {
+	if tp[i].Body.TeamType == "system" && tp[j].Body.TeamType == "system" {
+		return primitive.SystemTeams[tp[i].Body.Name] < primitive.SystemTeams[tp[j].Body.Name]
+	} else if tp[i].Body.TeamType == "system" {
+		return true
+	} else if tp[j].Body.TeamType == "system" {
+		return false
+	}
+
+	return tp[i].Body.Name[0] < tp[j].Body.Name[0]
+}
+
+func orgsMembersListCmd(ctx *cli.Context) error {
 
 	cfg, err := config.LoadConfig()
 	if err != nil {
@@ -248,10 +267,6 @@ func orgsMembersListCmd(ctx *cli.Context) error {
 
 	var org *envelope.Org
 	var orgs []envelope.Org
-	var team envelope.Team
-	var teams []envelope.Team
-	var memberships []envelope.Membership
-	var oErr, tErr, mErr, sErr error
 
 	// Retrieve the org name supplied via the --org flag.
 	// This flag is optional. If none was supplied, then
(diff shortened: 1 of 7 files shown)


cmd/orgs.go
```diff
@@ -215,6 +233,148 @@ func orgsRemove(ctx *cli.Context) error {
...
+		if oErr != nil {
+			return oErr
+		}
+
+		idx, _, oErr := SelectOrgPrompt(orgs)
+		if oErr != nil {
+			return oErr
+		}
+
+		if idx == promptui.SelectedAdd {
```
ianlivingstone: One thing we should think from a ux flow is whether or not we should allow you to create an org through the `torus orgs members` list. It seems disjointed that a user would ever want to create an org when they're listing members as the result would only include themselves.

On the other hand it fits into the way the rest of the product works.

cmd/orgs.go
```diff
@@ -44,6 +51,17 @@ func init() {
...
 			},
+			{
+				Name:  "members",
+				Usage: "List all members in an org",
+				Flags: []cli.Flag{
+					orgFlag("Use this organization.", false),
+				},
+				Action: chain(
+					ensureDaemon, ensureSession, loadDirPrefs, loadPrefDefaults,
+					setUserEnv, checkRequiredFlags, orgsMembersListCmd,
```
ianlivingstone: I don't think we need to `setUserEnv` or `checkRequiredFlags` here as the `orgFlag` is not required and we don't use the environment flag at all.

cmd/orgs.go
```diff
@@ -215,6 +233,148 @@ func orgsRemove(ctx *cli.Context) error {
...
+		return err
+	}
+
+	client := api.NewClient(cfg)
+	c := context.Background()
+
+	var getMembers sync.WaitGroup
+	getMembers.Add(2)
+
+	var org *envelope.Org
```
ianlivingstone: A small stylistic nitpick; it's easier to read if the variables are declared as close as possible to where they are used.

cmd/orgs.go
```diff
@@ -215,6 +233,148 @@ func orgsRemove(ctx *cli.Context) error {
...
+	var team envelope.Team
+	var teams []envelope.Team
+	var memberships []envelope.Membership
+	var oErr, tErr, mErr, sErr error
+
+	// Retrieve the org name supplied via the --org flag.
+	// This flag is optional. If none was supplied, then
+	// orgFlagArgument will be set to "". In this case,
+	// prompt the user to select an org.
+	orgFlagArgument := ctx.String("org")
```
ianlivingstone: Perhaps call this `orgName` as that's what would be contained within the variable instead of where the value came from?

cmd/orgs.go
```diff
@@ -215,6 +233,148 @@ func orgsRemove(ctx *cli.Context) error {
...
+		// Retrieve the team by name supplied
+		teams, tErr = client.Teams.GetByName(c, org.ID, teamName)
+		if len(teams) != 1 {
+			tErr = errs.NewExitError("Team not found.")
+			getMembers.Done()
+			return
+		}
+		team = teams[0]
+
+		// Hide machine teams from the teams list; as we use them to represent
```
ianlivingstone: You don't need this bit of code here as you are always looking up the `members` team

cmd/orgs.go
```diff
@@ -215,6 +233,148 @@ func orgsRemove(ctx *cli.Context) error {
...
+
+		if idx == promptui.SelectedAdd {
+			// COME BACK TO THIS
+			// Error condition, unsure how to handle
+		}
+
+		org = &orgs[idx]
+	}
+
+	go func() {
```
ianlivingstone: One thought I had (and totally  up for discussion) is whether the cognitive overhead of doing the lookup for the session concurrently with the lookup for org, team, and members is worth it. The session is actually cached inside the daemon, so it should be very fast anyhow!

ianlivingstone: It would drastically simplify this code as we wouldn't need the waitgroup or handle the multi error scenario.

cmd/orgs.go
ianlivingstone: We can also dry up this code a bit too, as this and the team members command are doing a lot of the same work.

cmd/orgs.go
MattWright481516: Agreed, this is my mistake. Looked too quickly at SelectCreateOrg which calls SelectOrgPrompt, thought SelectOrgPrompt presented the orgs without allowing creation of a new org.  Could make a new prmpt using Select instead of SelectWithAdd to disallow creation of a new org?

cmd/orgs.go
MattWright481516: I agree with avoiding concurrency here, I think the gain probably doesn't outweigh the confusion and the error handling is ugly. Also, in the event that no flag is used, there is a blocking API call anyway, followed up by the Go routines afterward, so I'm awkwardly mixing paradigms here. This was based on a similar function in teams.go that used the concurrent approach but to simplify things I'm happy to remove it.

cmd/orgs.go
MattWright481516: You're right, that was clumsy, will remove. Also agree that I'm duplicating some code from `teams members` so after this `orgs` cmd is working, can re-factor to avoid that.

cmd/orgs.go
ianlivingstone: Hmm, I thought a little bit more about this and while it's surprising its definitely not surprising given the way the rest of the CLI works. So I could go either way, but adding a new one for just selecting makes a lot of sense, or perhaps we open an issue and tackle it later.

cmd/orgs.go
```diff
@@ -215,6 +232,218 @@ func orgsRemove(ctx *cli.Context) error {
...
+	membershipUserIDs := make(map[identity.ID]bool)    // Set of unique user IDs
+
+	// Create:
+	//	- Set of unqiue user IDs in membershipUserIDs
+	// 	- Mapping from user IDs to team IDs in userTeamIdx (1:m mapping)
+	for _, membership := range memberships {
+		// Skip memberships not associated with teams within org
+		team, ok := teamsIdx[*membership.Body.TeamID]
+		if !ok {
+			continue
```
ianlivingstone: We should probably panic here as this shouldn't be possible (a membership existing for a team that doesnt exist). If it does happen then we should get alerted!

cmd/orgs.go
```diff
@@ -215,6 +232,218 @@ func orgsRemove(ctx *cli.Context) error {
...
+	for id := range membershipUserIDs {
+		userIDs = append(userIDs, id)
+	}
+
+	users, err := client.Profiles.ListByID(c, userIDs)
+	if err != nil {
+		return err
+	}
+	if users == nil {
+		return errs.NewExitError("User not found.")
```
ianlivingstone: Some errors you start with capital and others you don't, we should probably standardize them (e.g. `org not found` vs. `User not found.`)

cmd/orgs.go
```diff
@@ -215,6 +232,218 @@ func orgsRemove(ctx *cli.Context) error {
...
+
+func (tp ByTeamPrecedence) Len() int {
+	return len(tp)
+}
+
+func (tp ByTeamPrecedence) Swap(i, j int) {
+	tp[i], tp[j] = tp[j], tp[i]
+}
+
+func (tp ByTeamPrecedence) Less(i, j int) bool {
```
ianlivingstone: If you want, we could write a test to ensure the sorting works as expected.

ianlivingstone:
The other bit I forgot is that we need to write docs for the command as well!

Issue #302, opened 2017-11-02
Secret Retrieval Performance


ianlivingstone:
We've had a few people reach out to ask about performance improvements related to speeding up `torus run` and credential retrieval in general. As the number of unset and set credentials increases inside a path retrieval slows down.

I took a quick stab last night looking at what could be done today to improve retrieval. I've put a short list together below, but none of this is backed by any data based on benchmarking!

- [x] ~Skip unboxing of unset credentials, we only need to unbox always unbox v1 credentials as the fact they're unset is stored inside the CredentialValue instead of on the `state` field (which was added in v2 of the object).~ (Already done inside the daemon)
- [x] Pull sorting through unset and set credentials into the Daemon. Right now it lives inside the CLI which means we decrypt every single credential object inside a keyring and send the contents to the CLI or any other client. We can speed this up by moving the responsibility for sorting through set/unset into the Daemon and never sending unset to the CLI. In future we could add a query parameter to the retrieval route for including unset credentials if they're needed for a future command.
- [x] Break the unboxer into two parts, one which unboxes a Private Key, and another for actually unboxing a Keyring. This way we can group decryption by public key. The implication here is that most keyring memberships leverage the same public key for a user, so this should drastically reduce the number of private keys we need to unbox using the stored master key.
- [ ] ~Using go channels to parallelize decrypting different keyrings~ 
- [x] Parallelize retrieving the credentialgraph, claimtree and keypairs for a user

A few next steps:

- Get a profiler in place so it's easier to find the specific slowdowns/slow spots to determine where we're spending the most time.
- Get a few standard data sets in place to make it easy to benchmark as changes are made

PR #311, opened 2017-11-07
Separate unseal and unbox

ianlivingstone:
To remove duplication of unsealing of private keys, we've split `Unbox`ing into two explicit composable stages.

In order to leverage these changes, we've begun indexing graphs by the valid keyring membership used to include the targeted user/machine into the underlying ring.

Related #302

### DIFF
--- a/daemon/logic/engine.go
+++ b/daemon/logic/engine.go
@@ -262,8 +262,15 @@ func (e *Engine) RetrieveCredentials(ctx context.Context,
 	// actually do real work and decrypt each of these credentials but for
 	// now we just need ot return a list of them!
 	creds := []PlaintextCredentialEnvelope{}
-	for _, graph := range activeGraphs {
-		orgID := graph.GetKeyring().OrgID()
+	idx := newCredentialGraphKeyIndex(*(e.session.AuthID()))
+	idx.Add(activeGraphs...)
+
+	for encryptingKeyID, graphs := range idx.GetIndex() {
+		if len(graphs) == 0 {
+			continue
+		}
+
+		orgID := graphs[0].GetKeyring().OrgID()
 		kp, ok := keypairs[*orgID]
 		if !ok {
 			_, _, kp, err = fetchKeyPairs(ctx, e.client, orgID)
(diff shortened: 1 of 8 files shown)


daemon/logic/engine.go
```diff
@@ -274,64 +281,74 @@ func (e *Engine) RetrieveCredentials(ctx context.Context,
...
 				return nil, err
 			}
-			encryptingKeys[*krm.EncryptingKeyID] = encryptingKey
+			encryptingKeys[encryptingKeyID] = encryptingKey
 		}
 
-		err = e.crypto.WithUnboxer(ctx, *mekshare.Key.Value, *mekshare.Key.Nonce, &kp.Encryption, *encryptingKey.Key.Value, func(u crypto.Unboxer) error {
-			for _, cred := range graph.GetCredentials() {
-				pt, err := u.Unbox(ctx, *cred.Credential().Value, *cred.Nonce(), *cred.Credential().Nonce)
+		err = e.crypto.WithUnsealer(ctx, &kp.Encryption, *encryptingKey.Key.Value, func(unsealer crypto.Unsealer) error {
```
ianlivingstone: One downside is that we actually have the unsealed private key sitting around for longer inside memory. Something to think about.

registry/keyring.go
```diff
@@ -27,6 +27,7 @@ type KeyringSection interface {
 	GetKeyring() envelope.KeyringInf
 	KeyringVersion() int
 	FindMember(*identity.ID) (*primitive.KeyringMember, *primitive.MEKShare, error)
+	FindMemberByKeyID(*identity.ID) (*primitive.KeyringMember, *primitive.MEKShare, error)
```
ianlivingstone: This is really just looking for the `MEKShare` for a specific user, perhaps rename to `FindMEKShare` and make it explicit that it only accepts a PublicKeyID.

Furthermore, it should error if it can't find the MEKShare for the given ID or if the user's don't match? Or perhaps this is the wrong area to do that checking (the mekshare isn't sent from the server anyhow).

PR #323, opened 2017-11-29
Leverage memguard to securely store keys and tokens

ianlivingstone:
Instead of storing long lived encryption keys in standard memory which
can be swapped to disk or core dumped sensitive data is stored inside
guarded memory using the excellent golang
[memguard](github.com/awnumar/memguard) library.

Some values may be stored in non-secure memory during signup or login.

### DIFF
--- a/daemon/crypto/secure/secure.go
+++ b/daemon/crypto/secure/secure.go
@@ -1,0 +1,138 @@
+// Package secure is a wrapper package around memguard for managing data which
+// must not be swapped to disk or easily read via a memory scanner.
+package secure
+
+import (
+	"errors"
+	"sync"
+
+	"github.com/awnumar/memguard"
+)
+
+// ErrSecretNotFound occurs when a secret could not be found
+var ErrSecretNotFound = errors.New("Could not find secret to remove it")
+
+var lock *sync.Mutex
+var current *Guard
+
+func init() {
+	lock = &sync.Mutex{}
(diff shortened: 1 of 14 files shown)


daemon/crypto/secure/secure.go
```diff
@@ -0,0 +1,138 @@
...
+	if s == nil {
+		panic("received a nil secret; should not be possible")
+	}
+
+	g.lock.Lock()
+	defer g.lock.Unlock()
+
+	placement := -1
+	for i, v := range g.secrets {
+		if *v == *s {
```
awnumar: Did you mean to dereference `v` and `s` here? Instead of comparing the addresses of the structs, the contents are being compared here (which is not a huge deal *in this case*, but it seems unnecessary).

ianlivingstone: Great catch; what I really want here is to find if the secret points to the same underlying LockedBuffer instance. I'll open a pr with a fix!

awnumar: @ianlivingstone Ah, so you want to do `v.buffer == s.buffer`. But assuming that only a single secret holds a specific LockedBuffer instance, you should get away with `v == s`.

ianlivingstone: Yep! Just opened a PR with the change: https://github.com/manifoldco/torus-cli/pull/326

Thanks again for the catch :)

PR #328, opened 2017-12-01
Torus list

### LINKED ISSUE #322: Torus List Command
One feature that users frequently request is the ability to quickly view all secrets within an organization and/or project. This is crucial information that users should be able to view conveniently.

**Problem**

Specifically, the problems we would like to solve include:
- There is no single/easily readable list of all secrets stored within Torus and so this information is difficult to obtain.
- It is difficult to find out who has access to those secrets and vice versa, ie. which secrets a user or team has access to.
- It is difficult to understand/see the Torus project structure/hierarchy
- It is hard or impossible to search for secrets (ie. show all PORTs)

**Issues With Current Implementation**

The current solutions, ie. `torus ls` addresses some of these issues either directly or indirectly but suffers from its own issues, including:
- It performs unnecessary crypto in the background which slows it down.
- It is confusing and difficult to read/understand, and the output style is clunky.
- It is directly tied to the credential path. This means that instead of making the project structure obvious/readable, it requires the user to have an accurate mental model of the credential path already.

**Proposed Solution**

The solution proposed is to introduce a new command `torus list`. This command will list all secrets within a specified org/project in a style that makes the project structure apparent to the user. The user can pass any of several flags that will help filter out/narrow in on particular secrets. In general, this might look like:

```
torus list --org ORG --project PROJ [--environment ENV] [--service SERVICE] [--user USERNAME] [--team TEAMNAME] [--machine MACHINENAME] [--name SECRETNAME] [--verbose]
```

where

- `--org` is a non-optional flag specifying the organization name
- `--project` isa non-optional flag specifying the project name
- `--environment` is an optional flag that filters out all secrets except those within the environment (or environment expression)
- `--service` is an optional flag that filters out all secrets except those within the service (or service expression)
- `--user` is an optional flag that presents the intersection of all secrets that both the specified user and currently logged-in user have access to. By default, `torus list` necessarily filters out all secrets except those which the currently logged-in user has access to.
- `--team` is an optional flag that filters out all teams except the team specified.
- `--machine` is an optional flag the filters out all teams except the machine teams specified.
- `--name` is an optional flag that filters out all secrets except those with the specified name.
- `--verbose` is an optional flag that causes the full credential path of each secret to be listed.

**Example**

```
$torus list -o matt -p matts-project -e dev-* -s api

/matt/matts-project
        /dev-matt
                /api
                        port
                        password
        /dev-ian
                /api
                        port
                        password                        
```

**Notes**

One issue not addressed by this approach is an easy way to discover all the teams and users that have access to a particular secret (effectively the inverse of the currently proposed `torus list`). A solution to this problem requires an entirely different output style and set of manipulations and is best addressed with a separate command, eg. `torus inspect --secret`.

MattAWright:
Closes #322


cmd/list.go
```diff
@@ -0,0 +1,280 @@
...
+		Category:  "SECRETS",
+		Flags: []cli.Flag{
+			orgFlag("Use this organization.", false),
+			projectFlag("Use this project.", false),
+			envSliceFlag("Use this environment.", false),
+			serviceSliceFlag("Use this service.", "", false),
+			nameFlag("Find secrets with this name."),
+			cli.BoolFlag{
+				Name:  "verbose, v",
+				Usage: "Lists the sources of the secrets (shortcut for --format verbose)",
```
ianlivingstone: There is no format flag here, so this should be updated

cmd/list.go
```diff
@@ -0,0 +1,280 @@
...
+		Name:      "list",
+		ArgsUsage: "",
+		Usage:     "List your org's project structure and its secrets.",
+		Category:  "SECRETS",
+		Flags: []cli.Flag{
+			orgFlag("Use this organization.", false),
+			projectFlag("Use this project.", false),
+			envSliceFlag("Use this environment.", false),
+			serviceSliceFlag("Use this service.", "", false),
+			nameFlag("Find secrets with this name."),
```
ianlivingstone: Can you explain your reason for a name flag versus using positional arguments?

cmd/list.go
```diff
@@ -0,0 +1,280 @@
...
+	"github.com/manifoldco/torus-cli/envelope"
+	"github.com/manifoldco/torus-cli/errs"
+	"github.com/manifoldco/torus-cli/pathexp"
+)
+
+func init() {
+	list := cli.Command{
+		Name:      "list",
+		ArgsUsage: "",
+		Usage:     "List your org's project structure and its secrets.",
```
ianlivingstone: Perhaps something along the lines of `List allows you to list and filter all secrets you can access inside a project` ?

cmd/list.go
ianlivingstone: Actually, it looks like this was forgotten?

cmd/list.go
```diff
@@ -0,0 +1,280 @@
...
+
+	cfg, err := config.LoadConfig()
+	if err != nil {
+		return err
+	}
+
+	client := api.NewClient(cfg)
+	c := context.Background()
+
+	// TODO - Make this a type
```
ianlivingstone: Is this a todo you're going to resolve ?

cmd/list.go
```diff
@@ -0,0 +1,280 @@
...
+		}
+	}
+
+	projectName := ctx.String("project")
+	var project *envelope.Project
+	if projectName == "" {
+		// Retrieve list of available projects
+		projects, err := client.Projects.List(c, org.ID)
+		if err != nil {
+			return errs.NewExitError("Failed to retrieve projects list.")
```
ianlivingstone: We shoudl probably use `NewErrorExitError` here and in other places so we actually return the underlying error, we don't know why it failed!

cmd/list.go
```diff
@@ -0,0 +1,280 @@
...
+
+	// Retrieve service flag values
+	// If no values were set, use a full glob
+	serviceFilters := ctx.StringSlice("service")
+	if len(serviceFilters) == 0 {
+		serviceFilters = append(serviceFilters, "*")
+	}
+
+	// The following two slices are placeholders necessary to
+	// build the PathExp later. Check that this is true, if I pass
```
ianlivingstone: Did you answer your question you left here?

cmd/list.go
```diff
@@ -0,0 +1,280 @@
...
+
+	// Check for each env and service in the filtered list, add
+	// any credentials along that path to that env/service branch
+	// of the credentialsTree
+	projectPath := "/" + orgName + "/" + projectName + "/"
+	for _, e := range filteredEnvNames {
+		for _, s := range filteredServiceNames {
+			builtPathExp, err := pathexp.Parse(projectPath + e + "/" + s + "/*/*")
+			for _, cred := range credentials {
+				if len(args) > 0 && isSecretNameInList((*cred.Body).GetName(), args) == false {
```
ianlivingstone: you could simplify this to `if len(args) > 0 && !isSecretNameInList()`

cmd/list.go
```diff
@@ -0,0 +1,280 @@
...
+					continue
+				}
+				credPathExp := (*cred.Body).GetPathExp()
```
ianlivingstone: You're also derferencing the `cred.Body` frequerntly you could just store it as a var then re-use it (e.g. `body :=  *cred.Body`)

cmd/list.go
```diff
@@ -0,0 +1,280 @@
...
+				if err != nil {
```
ianlivingstone: This should be on L225?

cmd/list.go
```diff
@@ -0,0 +1,280 @@
...
+					return errs.NewErrorExitError("Failed to parse: "+projectPath+e+"/"+s+"/*/*", err)
+				}
+				// If cred not contained in any builtPathExps, it is not
+				// within the search space specified by the flags.
+				if credPathExp.Contains(builtPathExp) == false {
```
ianlivingstone: as above, this oculd be simplified: `!credPathExp.Contains(builtPathExp)`

cmd/list.go
```diff
@@ -0,0 +1,280 @@
...
+	fmt.Println("")
+	w := tabwriter.NewWriter(os.Stdout, 5, 0, 2, ' ', 0)
+	fmt.Fprintf(w, "%s\n", projectPath)
+	for e := range credentialsTree {
+		fmt.Fprintf(w, "\t%s\t\n", e+"/")
+		for s := range credentialsTree[e] {
+			fmt.Fprintf(w, "\t\t%s\t\n", s+"/")
+			if len(credentialsTree[e][s]) == 0 && verbose {
+				fmt.Fprintf(w, "\t\t\t[empty]\n")
+			} else {
```
ianlivingstone: You could simplify this using `continue`:

```
if len(credentialsTree[e][s]) == 0 {
  if verbose {
    fmt.Fprintf(w, "\t\t\t[empty]\n")
  }

  continue
}
```

docs/commands/secrets.md
```diff
@@ -187,56 +187,71 @@ To ensure that your command’s arguments and options are passed correctly you m
...
 ### Command Options
 
   Option | Description
   ---- | ----
+  --org, -o | Required flag to specify org.
+  --project, -p | Required flag to specify project.
+  --env, -e | Specify environment filter(s) for displayed secrets. To specify multiple environments, pass multiple flags (eg.`torus list -e env1 -e env2`). This flag is optional.
+  --service, -s | Specify service filter(s) for displayed secrets. To specify multiple services, pass multiple flags (eg. `torus list -s ser1 -s ser2`). This flag is optional.
   --verbose, -v | Show which type of path is being displayed, shortcut for --format=verbose
   --format FORMAT, -f FORMAT | Format used to display data (simple, verbose) (default: simple)
```
ianlivingstone: There is no format flag!

docs/commands/secrets.md
```diff
@@ -187,56 +187,71 @@ To ensure that your command’s arguments and options are passed correctly you m
...
 
 ### Command Options
 
   Option | Description
   ---- | ----
+  --org, -o | Required flag to specify org.
+  --project, -p | Required flag to specify project.
+  --env, -e | Specify environment filter(s) for displayed secrets. To specify multiple environments, pass multiple flags (eg.`torus list -e env1 -e env2`). This flag is optional.
+  --service, -s | Specify service filter(s) for displayed secrets. To specify multiple services, pass multiple flags (eg. `torus list -s ser1 -s ser2`). This flag is optional.
   --verbose, -v | Show which type of path is being displayed, shortcut for --format=verbose
```
ianlivingstone: also, there is no format flag!

docs/commands/secrets.md
```diff
@@ -187,56 +187,71 @@ To ensure that your command’s arguments and options are passed correctly you m
...
 
-## ls
-###### Added [v0.13.0](https://github.com/manifoldco/torus-cli/blob/master/CHANGELOG.md)
-
-`torus ls [path]` allows a user to explore the objects and values set within their organization.
+## list
+###### Added [v0.28.0](https://github.com/manifoldco/torus-cli/blob/master/CHANGELOG.md)
 
-Each level within the organization can be inspected by changing the segments supplied in the path. Wildcards cannot be used for the organization or project segments of the [path](../concepts/path.md).
+`torus list [name]...` allows a user to explore the objects and values set within their organization. Users can find the location of a particular secret by passing its name as an argument, or view all secrets within different branches of their organization.
```
ianlivingstone: A suggestion for better clarity:

```
`torus list [name]...` allows a user to explore the secrets stored within a project. Users can search for specific secrets by supply a name as an argument or filter down to specific environments and servers using the `--environment, e` and `--service, s` flags.
```

docs/commands/secrets.md
```diff
@@ -187,56 +187,71 @@ To ensure that your command’s arguments and options are passed correctly you m
...
 
-Path is required, and does not support context.
+Each level within the organization can be inspected by changing the options supplied to the command.
```
ianlivingstone: If you were to take my suggestion, I think this sentence would be redundant.

docs/commands/secrets.md
```diff
@@ -187,56 +187,71 @@ To ensure that your command’s arguments and options are passed correctly you m
...
+            secret1
 ```
 
-Expanding the path to view secrets using recursive:
+List all secrets in a specific service:
 ```
-$ torus ls /my-org/landing-page/dev-jeff -r
-/my-org/landing-page/dev-*/[api|www]/*/*/port
-/my-org/landing-page/[dev-jeff|dev-sally]/www/*/*/token
+$ torus list -s ser1
```
ianlivingstone: An example showing `-v` would be good as well, along with passing more than one service and environment, or using a wildcard!

cmd/list.go
MattAWright: Woops, yeah it was forgotten. Leftover from an early implementation that used flags instead of arguments.

cmd/list.go
MattAWright: Good catch

cmd/list.go
MattAWright: Yes, have resolved the levels of maps into seperate types, going into the next push.

cmd/list.go
MattAWright: Good catch, missed that.

cmd/list.go
MattAWright: Yep, original code is right and necessary.

cmd/list.go
MattAWright: Yes, always a bit of an internal debate. Nice variable name at the cost of an extra line to assign it, or save the line in a dense spot but with uglier naming. Switched to the added line, probably best.

cmd/list.go
MattAWright: Woops!

cmd/list.go
MattAWright: Yes, good thought, fixed

docs/commands/secrets.md
MattAWright: Good catch, second time I've missed a ref to that.

docs/commands/secrets.md
MattAWright: Fair, added

docs/commands/secrets.md
MattAWright: Same as above

docs/commands/secrets.md
MattAWright: Yep, you're right.

docs/commands/secrets.md
MattAWright: Fair, made change, fixed typos.

cmd/list.go
ianlivingstone: Looks like you forgot to remove the comment!

MattAWright: Fixed.

PR #332, opened 2017-12-07
Use dep

### LINKED ISSUE #300: Switch from Glide to Godep
It's currently planned for godep to become a part of the standard golang tool set in `go 1.10`. We could reduce a dependency by switching off on glide to godep.

This will also give us the ability to use the very wonderful `dep ensure` command!

arbourd:
Switch to [golang/dep](https://github.com/golang/dep)

Some between dep and glide:
- Cannot ignore `golang.org/x/sys/unix` (build fails as it doesn't import)
- dep vendors all dependencies when running `dep ensure` ([for the time being](https://github.com/golang/dep/issues/944)) -- use `dep prune` to clean up unused dependencies
- unix builds are using [dep at master](https://github.com/manifoldco/torus-cli/pull/332/commits/34c8071060d0b4b817efc19a85038ce89071f617#diff-b67911656ef5d18c4ae36cb6741b7965R37) -- i can download the binary in the Makefile instead

Closes #300


ianlivingstone:
This is awesome, thanks so much for the contribution!

arbourd:
My pleasure. 🤓

Just a quick note: `go list ./... | grep -v ./vendor` shouldn't be need with Go 1.9+ which automatically ignores the vendor folder.

PR #333, opened 2017-12-07
Named pipe security attributes

morochena:
Addresses issue #215

### DIFF
--- a/daemon/socket/socket_windows.go
+++ b/daemon/socket/socket_windows.go
@@ -4,6 +4,7 @@ package socket
 
 import (
 	"fmt"
+	"log"
 	"net"
 	"os/user"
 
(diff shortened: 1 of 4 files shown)


daemon/socket/socket_windows.go
```diff
@@ -3,26 +3,29 @@
...
-	// named pipe gets a default security descriptor and the handle cannot be
-	// inherited. The ACLs in the default security descriptor for a named pipe
-	// grant full control to the LocalSystem account, administrators, and the
-	// creator owner. They also grant read access to members of the Everyone
-	// group and the anonymous account.
-	return npipe.Listen(transportAddress)
+	// Gets current user's SID 
+	usr, err := user.Current()
+	if err != nil {
+		fmt.Printf("Error getting user SID: %s\n", transportAddress)
```
ianlivingstone: Instead of using `fmt` here, we should try to use the `log` package

PR #334, opened 2017-12-07
MSI Installer

morochena:
Uses go-msi to generate a .msi installer for windows, and allows for distribution via chocolately.


appveyor.yml
```diff
@@ -26,6 +33,8 @@ install:
 
 build_script:
   - c:\gopath\bin\dep.exe ensure
+  - go build -o torus.exe
```
ianlivingstone: Hmm, we're probably going to have to separate out building the windows binary (and now the installer + uploading it) from the rest of the Makefile when we build a release through `make release-all`.

If you want to introduce a `make msi` that requires the os to be windows and for `wix/go-msi` to be available (if possible) then I think we can open another issue (or subsequent PR) that has the changes to actually separately build, package, and upload a windows release.

packaging/msi/templates/product.wxs
```diff
@@ -0,0 +1,150 @@
...
+
+      <Directory Id="TARGETDIR" Name="SourceDir">
+
+         <Directory Id="$(var.Program_Files)">
+            <Directory Id="INSTALLDIR" Name="{{.Product}}">
+               {{if gt (.Files.Items | len) 0}}
+               <Component Id="ApplicationFiles" Guid="{{.Files.GUID}}">
+                  {{range $i, $e := .Files.Items}}
+                    <File Id="ApplicationFile{{$i}}" Source="{{$e}}"/>
+                    <!--
```
ianlivingstone: We can remove this right? Since we're not installing it as a service?

packaging/msi/templates/product.wxs
```diff
@@ -0,0 +1,150 @@
...
+            Name="{{.Product}}"
+            Version="{{.VersionOk}}"
+            Manufacturer="{{.Company}}"
+            Language="1033">
+
+      <Package InstallerVersion="200" Compressed="yes" Comments="Windows Installer Package" InstallScope="perMachine"/>
+
+      <Media Id="1" Cabinet="product.cab" EmbedCab="yes"/>
+
+      <Upgrade Id="{{.UpgradeCode}}">
```
ianlivingstone: This is just a rule for forcing the upgrade of Torus right? Is there a way we can run a command at the same time?

When we upgrade an existing install we should also restart the daemon via `torus daemon stop`. Ideally we also tell them that after upgrading they will need to log into Torus again via the command prompt (or whichever shell they use).

wix.json
```diff
@@ -0,0 +1,27 @@
+{
+  "product": "torus",
+  "company": "manifoldco",
+  "license": "LICENSE.md",
+  "version": "0.0.1",
```
ianlivingstone: This is the version of the binary right? Ideally this file can be tempalted like `package.json` inside `packaging/npm` which is then filled in prior to packaging.

This way we can pull the correct version in as well.

wix.json
```diff
@@ -0,0 +1,27 @@
...
+  "upgrade-code": "8615055C-D8E0-404C-93BE-441C503BA6F0",
```
ianlivingstone: What sets this upgrade code?

packaging/msi/templates/choco/LICENSE.txt
```diff
@@ -0,0 +1,9 @@
+From: {{.Choco.LicenseURL}}
```
ianlivingstone: Thinking a bit about this, for now it's probably easiest just to get the wix support into the build pipeline so we're producing and uploading a real msi installer, then come back to thinking about Choco support.

Thoughts?

appveyor.yml
```diff
@@ -14,6 +14,13 @@ init:
   - git config --global core.autocrlf input
 
 install:
+  # wix setup
+  - set PATH=%WIX%\bin;%PATH%
+  # go-msi install - static link method
```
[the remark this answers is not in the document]
morochena: Agreed, I'll add mingw32-make to appveyor so we're able to use the Makefile in appveyor.yml.

packaging/msi/templates/choco/LICENSE.txt
morochena: Sounds good, I'll remove the Chocolatey artifacts for now.

appveyor.yml
morochena: I introduced an initial `make msi` that checks for `go-msi`, but I'll add os detection and hook it into the appveyor.yml.

wix.json
morochena: `$ go-msi set-guid` generates them

packaging/msi/templates/choco/pkg.nuspec
```diff
@@ -0,0 +1,38 @@
+<?xml version="1.0"?>
```
ianlivingstone: We don't need these templates for `choco` right?

Makefile
```diff
@@ -454,3 +454,38 @@ builds/torus-npm-$(VERSION).tar.gz: npm
...
+	curl --create-dirs -fsSL -k -o C:\Temp\go-msi.msi http://github.com/mh-cbon/go-msi/releases/download/1.0.2/go-msi-amd64.msi
+	msiexec /i C:\Temp\go-msi.msi /q
+endif
+.PHONY: bootstrap-windows
+	
+#################################################
+# Generate Windows .msi 
+#################################################
+
+msi: binary
```
ianlivingstone: I think this should depend on the `release-binary-windows` rule? This way it also gets put inside the `builds/` directory. The subsequent msi should probably also be placed there as well, so it gets deleted on `make clean`.

wix.json
```diff
@@ -0,0 +1,30 @@
+{
```
ianlivingstone: This shouldn't be checked in right as it's a templated file?

PR #341, opened 2017-12-20
Colors

MattAWright:


cmd/list.go
```diff
@@ -5,16 +5,18 @@ import (
 	"fmt"
 	"os"
 	"sync"
-	"text/tabwriter"
+	//"text/tabwriter"
```
ianlivingstone: forgotten?

cmd/envs.go
```diff
@@ -163,69 +153,92 @@ func listEnvsCmd(ctx *cli.Context) error {
...
-	}
-	if org == nil {
-		return errs.NewExitError("Org not found.")
+	var orgs []envelope.Org
+
+	// Retrieve the org name supplied via the --org flag.
+	// This flag is optional. If none was supplied, then
+	// orgFlagArgument will be set to "". In this case,
+	// prompt the user to select an org.
+	orgName := ctx.String("org")
```
ianlivingstone: Could we dry this up? We repeat it in several different places.

cmd/invites.go
```diff
@@ -25,9 +25,9 @@ func init() {
 				Name:  "list",
 				Usage: "List outstanding invitations for an organization. These invites have yet to be approved.",
 				Flags: []cli.Flag{
-					orgFlag("org to list invites for", true),
+					orgFlag("org to list invites for", false),
 					cli.BoolFlag{
-						Name:  "approved",
+						Name:  "all, a",
 						Usage: "Show only approved invites",
```
ianlivingstone: This usage isn't correct anymore, is it? If so can you update it? We didn't actually update the behaviour either, it shows only approved invites. 

So, if it's going to be `--all` it should show all invites, if not, it should probably continue being `--approved`, the shortcut `-a` is 👍  either way!

cmd/invites_list.go
```diff
@@ -4,16 +4,19 @@ import (
 	"context"
 	"fmt"
 	"os"
-	"text/tabwriter"
+	//"text/tabwriter"
```
ianlivingstone: Forgotten?

cmd/orgs.go
```diff
@@ -7,9 +7,10 @@ import (
 	"sort"
 	"strconv"
 	"sync"
-	"text/tabwriter"
+	//"text/tabwriter"
```
ianlivingstone: forgotten?

cmd/projects.go
```diff
@@ -53,22 +53,67 @@ func init() {
...
-	fmt.Println(strings.Repeat("-", utf8.RuneCountInString(title)))
+	fmt.Printf("  %s\n", ui.Bold("Projects"))
 	for _, project := range projects {
-		fmt.Println(project.Body.Name)
+		fmt.Printf("  %s\n", project.Body.Name)
 	}
 	fmt.Println("")
 
+	count := strconv.Itoa(len(projects))
+	countStr := "Org " + orgName + " has (" + count + ") projects."
```
ianlivingstone: You could do this with `fmt.Printf("Org "%s has (%d) projects.\n", orgName, len(projects))`

ui/color.go
```diff
@@ -0,0 +1,64 @@
...
+
+func Color(c ansiterm.Color, s string) string {
+  ctx := ansiterm.Context {
+    Foreground: c,
+  }
+
+  return createStyledString(ctx, s)
+}
+
+func Environment(s string) string {
```
ianlivingstone: Are we using these anywhere? if not, we should cut them loose.

errs/errs.go
```diff
@@ -37,3 +37,14 @@ func NewExitError(message string) error {
 	}
 	return cli.NewExitError(message, -1)
 }
+
+func FilterErrors(errors ...error) cli.MultiError {
```
ianlivingstone: Instead of `FilterErrors` perhaps this should just be `MultiError` as its a MultiError wrapper?

cmd/invites.go
MattAWright: Yes, this was clumsy, misunderstood the behaviour. Have reverted it bad to "approved" but added the short "a". After taking another look, being able to show all approved vs all other types makes sense.

cmd/envs.go
MattAWright: Done

cmd/invites_list.go
MattAWright: Yep, fixed

cmd/list.go
MattAWright: Yep

cmd/projects.go
MattAWright: Yeah now that you say this, it's weird I would ever do otherwise. This is everywhere so I'll change it in other places too.

PR #349, opened 2017-12-22
Colors

### LINKED ISSUE #346: Bug - torus list produces empty output if no secrets set
If a user doesn't have access to any secrets then `torus list` outputs nothing. We should probably return a nice message perhaps with an emoji for this default state.

We should probably also hint regarding `torus set`.

### LINKED ISSUE #347: Colourize torus policies
Unlike many of the other listing and viewing commands, `torus policies list` has not been colourized.

- [ ] `torus policies list`
- [ ] `torus policies view`

MattAWright:
Made some fixes after running through QA checklist. Fixed bug in project retrieval, added output to list when no secrets are available.

Fixes #346 
Fixes #348 
Fixes #347

### DIFF
--- a/cmd/list.go
+++ b/cmd/list.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"os"
+	"strconv"
 	"sync"
 
 	"github.com/urfave/cli"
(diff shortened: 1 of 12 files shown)


cmd/list.go
```diff
@@ -290,7 +245,8 @@ func listCmd(ctx *cli.Context) error {
 		}
 	}
 	w.Flush()
-	fmt.Println("")
+
+	fmt.Printf("\n(%d) secrets found\n.", credCount)
```
ianlivingstone: We could add some color to the number (light gray?)

cmd/list.go
```diff
@@ -262,8 +216,8 @@ func listCmd(ctx *cli.Context) error {
 	if(verbose){
 		fmt.Println("")
 		projW := ansiterm.NewTabWriter(os.Stdout, 0, 0, 4, ' ', 0)
-		fmt.Fprintf(projW, "Org:\t" + ui.Faint(orgName) + "\t\n")
-		fmt.Fprintf(projW, "Project:\t" + ui.Faint(projectName) + "\t\n")
+		fmt.Fprintf(projW, "Org:\t" + ui.Faint(org.Body.Name) + "\t\n")
```
ianlivingstone: We should probably bold org and project?

PR #353, opened 2018-01-08
Spinner

MattAWright:
Added Spinners to UI package and a spinner generator function to cmd package. Added spinners to: set, unset, invites accept, invites approve, keypairs generate and keypairs revoke, import, export, view

### DIFF
--- a/cmd/import.go
+++ b/cmd/import.go
@@ -61,10 +61,10 @@ func importCmd(ctx *cli.Context) error {
 	s, p := spinner("Attempting to set credentials...")
 	s.Start()
 	creds, err := setCredentials(ctx, path, makers, p)
+	s.Stop()
 	if err != nil {
 		return errs.NewErrorExitError("Could not set credentials.", err)
 	}
-	s.Stop()
 
 	fmt.Println()
 	for _, cred := range creds {
--- a/cmd/import.go
+++ b/cmd/import.go
@@ -8,10 +8,11 @@ import (
 	"strings"
 
 	"github.com/google/shlex"
+	"github.com/urfave/cli"
+
 	"github.com/manifoldco/torus-cli/apitypes"
 	"github.com/manifoldco/torus-cli/errs"
 	"github.com/manifoldco/torus-cli/hints"
-	"github.com/urfave/cli"
 )
 
 type secretPair struct {
@@ -57,10 +58,13 @@ func importCmd(ctx *cli.Context) error {
 		}(secret.value)
 	}
 
-	creds, err := setCredentials(ctx, path, makers)
+	s, p := spinner("Attempting to set credentials...")
+	s.Start()
+	creds, err := setCredentials(ctx, path, makers, p)
 	if err != nil {
 		return errs.NewErrorExitError("Could not set credentials.", err)
 	}
+	s.Stop()
 
 	fmt.Println()
 	for _, cred := range creds {
--- a/cmd/import.go
+++ b/cmd/import.go
@@ -58,7 +58,7 @@ func importCmd(ctx *cli.Context) error {
 		}(secret.value)
 	}
 
-	s, p := spinner("Attempting to set credentials...")
+	s, p := spinner("Attempting to set credentials")
 	s.Start()
 	creds, err := setCredentials(ctx, path, makers, p)
 	s.Stop()
--- a/cmd/invites_approve.go
+++ b/cmd/invites_approve.go
@@ -53,7 +53,10 @@ func invitesApprove(ctx *cli.Context) error {
 		return errs.NewExitError("Invite not found.")
 	}
 
-	err = client.OrgInvites.Approve(context.Background(), *targetInvite, progress)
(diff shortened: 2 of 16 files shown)


cmd/invites_approve.go
```diff
@@ -53,7 +53,10 @@ func invitesApprove(ctx *cli.Context) error {
 		return errs.NewExitError("Invite not found.")
 	}
 
-	err = client.OrgInvites.Approve(context.Background(), *targetInvite, progress)
+	s, p := spinner(fmt.Sprintf("Reviewing invite for %s", email))
```
ianlivingstone: I'm thinking the text here should say "Attempting to approve invite for %s" or similar?

cmd/import.go
```diff
@@ -57,7 +58,10 @@ func importCmd(ctx *cli.Context) error {
 		}(secret.value)
 	}
 
-	creds, err := setCredentials(ctx, path, makers)
+	s, p := spinner("Attempting to set credentials...")
```
ianlivingstone: In some places you're using `...` and others not, we should standardize the messages along with the ones coming back from the daemon.

ui/spinner.go
```diff
@@ -0,0 +1,47 @@
...
+
+  "github.com/briandowns/spinner"
+)
+
+type Spinner struct{
+  spinner *spinner.Spinner
+  text string
+}
+
+func NewSpinner(text string) *Spinner { return defUI.NewSpinner(text) }
```
ianlivingstone: I think this function and the one below should be in the `ui/ui.go` file as thats where the `UI` struct and subsequent global `defaultUI` are defined. Then this file would only contain the spinner struct itself and it's subsequent methods.

ui/spinner.go
```diff
@@ -0,0 +1,47 @@
...
+  s.Suffix = " " + text
+  return &Spinner {
+    s,
+    text,
+  }
+}
+
+func (s *Spinner) Start() { defUI.StartSpinner(s) }
+
+func (u *UI) StartSpinner(s *Spinner) {
```
ianlivingstone: Why is this method on the `UI` struct inside of the `Spinner` struct? I think it'd be cleaner/clearer lines of responsibility for the `Spinner` struct to have the values it needs off the `UI` struct copied onto itself (or just a reference to the `UI` struct in general).

ui/spinner.go
```diff
@@ -0,0 +1,47 @@
+package ui
+
+import (
+  "time"
+
+  "github.com/briandowns/spinner"
+)
+
+type Spinner struct{
```
ianlivingstone: Since these are public methods they'll require documentation strings to pass CI (gofmt)

ui/spinner.go
```diff
@@ -0,0 +1,47 @@
...
+func (u *UI) NewSpinner(text string) *Spinner {
+  s := spinner.New(spinner.CharSets[9], 100*time.Millisecond)
+  s.Suffix = " " + text
+  return &Spinner {
+    s,
+    text,
+  }
+}
+
+func (s *Spinner) Start() { defUI.StartSpinner(s) }
```
ianlivingstone: Start/stop should be checking whether or not stdout is attached to a terminal or not. I think, if we're not attached to a terminal, we shouldn't print out the spinners (just make it an empty do-nothing call like several other `ui` functions).

cmd/import.go
MattAWright: Agreed, missed this. Standard is no ellipses

ui/spinner.go
```diff
@@ -0,0 +1,37 @@
+package ui
```
ianlivingstone: The indentation in this file is wrong, should be 8 spaces

ui/ui.go
```diff
@@ -81,6 +81,26 @@ type UI struct {
...
+
+func (u *UI) StartSpinner(s *Spinner) {
+  if !u.EnableProgress || !readline.IsTerminal(int(os.Stdout.Fd())) {
+    return
+  }
+
+	s.Start()
+}
+
+func (u *UI) StopSpinner(s *Spinner) {
```
ianlivingstone: This method should be documented

ui/ui.go
```diff
@@ -81,6 +81,26 @@ type UI struct {
 	EnableColors bool
 }
 
+func (u *UI) NewSpinner(text string) *Spinner {
+	return NewSpinner(text)
+}
+
+func (u *UI) StartSpinner(s *Spinner) {
```
ianlivingstone: This method should be documented

ui/ui.go
```diff
@@ -81,6 +81,26 @@ type UI struct {
 	EnableColors bool
 }
 
+func (u *UI) NewSpinner(text string) *Spinner {
```
ianlivingstone: This method should be documented

ui/spinner.go
ianlivingstone: Also the struct should be documented as well

PR #357, opened 2018-01-23
Conversion to Promptui

### LINKED ISSUE #303: Switch to using manifoldco/promptui
PromptUI has been pulled out into a [separate repository](https://github.com/manifoldco/promptui) while adding features and tests.

We should look at switching over to the latest version of PromptUI and moving/resolving any outstanding bugs recorded inside this repository related to promptui.

ianlivingstone:
Instead of continuing to use and maintain our promptui library, we've
decided to switch to `manifoldco/promptui` which was pulled out from
torus-cli some time ago.

At the same time, I've added better support for "detached terminal"
states where stdout is not a terminal. In these situations, the user is
not prompted, instead, the provided value (if any) is validated and
checked against known options.

On top of these changes, I've added a wrapper around every command to
ensure that errors are displayed properly.

Several other cosmetic changes have been bundled in to make it easier to
display warnings and other types of information to the user as a part of
the `ui` package.

Closes #303


prompts/prompts.go
```diff
@@ -0,0 +1,185 @@
...
+
+	"github.com/manifoldco/torus-cli/errs"
+	"github.com/manifoldco/torus-cli/prefs"
+	"github.com/manifoldco/torus-cli/ui"
+	"github.com/manifoldco/torus-cli/validate"
+)
+
+// StringPrompt represents a function that prompts the user and returns a
+// string. It supports accepting a string as a default value and a boolean
+// representing whether it should automatically validates the data.
```
MattAWright: "automatically validates" --> "automatically validate"

PR #358, opened 2018-01-25
Add dep prune options to dep manifest

arbourd:
With 0.4.0, dep now runs `prune` automatically when when using `ensure`.
Dep will check in the manifest for what options are needed for pruning.

This commit sets the prune settings to remove non-go files (exlcuding
LICENSE and AUTHORS), go test files. Unused (non-imported) packages need
to be included because of the tools/ package.

### DIFF
--- a/Gopkg.toml
+++ b/Gopkg.toml
@@ -128,3 +128,7 @@ required = [
 [[constraint]]
   name = "github.com/google/go-github"
   revision = "466070b0580728e63bd1a415e0019639e55d7148"
+
+[prune]
+  non-go = true
+  go-tests = true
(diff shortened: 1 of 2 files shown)


arbourd:
Looks like the build is failing because this broke a cache. I'm seeing the same errors in master on my fork: https://travis-ci.org/arbourd/torus-cli/builds/333346471

arbourd:
Build failures were due to the `dep prune` settings. Removing unused dependencies pruned `go-bindata`, which is not imported, but needed for the toolchain.

Sorry for all the failures!

ianlivingstone:
Awesome, thanks for this :)

PR #361, opened 2018-01-30
Standardize list output format and coloring

ianlivingstone:
A few of our list commands did not output the number of items in the
list and others did not properly colorize the integer.

I've gone through and standardized these across Torus.

### DIFF
--- a/cmd/orgs.go
+++ b/cmd/orgs.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"os"
 	"sort"
+	"strconv"
 	"sync"
 
 	"github.com/juju/ansiterm"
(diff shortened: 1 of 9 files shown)


cmd/orgs.go
```diff
@@ -145,7 +146,8 @@ func orgsListCmd(ctx *cli.Context) error {
 		fmt.Printf("%s\n", o.Body.Name)
 	}
 
-	hints.Display(hints.PersonalOrg)
+	fmt.Printf("\nYou are a member of (%s) org%s\n",
```
MattAWright: No more personal org hint?

cmd/signup.go
```diff
@@ -80,7 +80,7 @@ func signup(ctx *cli.Context, subCommand bool) error {
 	}
 
 	reminderLabel := "Reminder: "
-	ui.Hint("Don't forget to keep your password safe and secure! You can't recover your account if you lossswordd.", false, &reminderLabel)
```
MattAWright: How did that even get in there?

ianlivingstone: The magic of random key mashing in vim

bitmoji

cmd/orgs.go
ianlivingstone: I removed it because it made the output really messy, would be fine with adding it back.

PR #364, opened 2018-02-02
Team filter

MattAWright:
Added a --team, -t flag to "list" command. This let's users filter base don what secrets a certain team or group of teams have access to. Leverages new ACL changes in the server.

### DIFF
--- a/docs/commands/secrets.md
+++ b/docs/commands/secrets.md
@@ -239,8 +239,9 @@ service: default
   ---- | ---- | ----
   --org, -o | TORUS_ORG | Required flag to specify org.
   --project, -p | TORUS_PROJECT | Required flag to specify project.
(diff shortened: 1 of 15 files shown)


docs/commands/secrets.md
```diff
@@ -322,3 +323,27 @@ env1/
     ser1/
         secret1   (/org/project/env1/ser1/secret1)
 ```
+
+**Use `--team` flag to limit the secrets displayed by list. In this example, the team "team1" only has access to the secrets in "staging"**
```
ianlivingstone: We should probably separate these into two lines, e.g.:

```
**List only secrets a specific team can access**

In this example, the team `team1` only has access to the secrets in the staging environment of the current project.
```

Issue #377, opened 2018-03-16
Decrypting Credentials


ItsLeeOwen:
Every nth (5th ish) time that torus run is executed, it hangs on Decrypting Credentials indefinitely until the machine is restarted, or background torus processes are force quit.

Are there any verbosity flags to help troubleshoot this?

ianlivingstone:
Yeah, if you watch the daemon log `$HOME/.torus/daemon.log`, you should be able to see the outgoing HTTP requests from the daemon.

I'm going to dig in real quick and see if there is an obvious issue or not.

briangruber:
Has this been resolved? Myself and several in the office seem to get this a lot and have to `torus daemon stop` or restart computer to get it off Decrypting Credentials. We are using v0.30.1

ItsLeeOwen:
I regularly experience the originally described issue when running multiple apps simultaneously with torus-cli 0.30.1 on OSX installed via homebrew.  Stopping the running process to relogin and resume is my current workaround.

PR #378, opened 2018-03-17
Spinners obey core.progress pref and unblock view

### LINKED ISSUE #376: CLI ignores `core.progress` flag
```
$ torus prefs list

/Users/radmen/.torusrc

[core]
    registry_uri       = https://registry.arigato.sh
    manifest_uri       = https://get.torus.sh/manifest.json
    gatekeeper_address = 0.0.0.0:8200
    context            = true
    auto_confirm       = false
    progress           = false
    hints              = false
    vim                = false
    check_updates      = true
    colors             = true

$ torus export -e test > .env
$ vim .env
```

Here's the result of export:

zrzut ekranu 2018-03-16 o 12 19 24

Unfortunately Torus ignored `core.progress` flag.

ianlivingstone:
We had expoed spinners through a few different apis in the ui layer, as
a result some were obeying the `core.progress` preference while others
were not. I've fixed this by pushing the responsibility for checking
whether a spinner should/is displayed into the actual spinner itself.

At the sametime, I've fixed a bug where it was possible for a spinner to
never be removed due to a panic.

Closes #376 #377

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,17 @@
 # CHANGELOG
 
+## Unreleased
+
+_YYYY-MM-DD_
+
+**Fixes**
+
+- The spinners released in `v0.30.0` can now be disabled by setting the
+  `core.progress` preference to false via `torus prefs set core.progress
+  false`.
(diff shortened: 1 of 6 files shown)


radmen:
Thank you!

